PR #342, opened 2026-04-17
feat(data): verified_scene_ids.json registry + loader + tests (A5.2)

cha-ndler:
## Summary

- Adds `src/main/resources/com/collectionloghelper/verified_scene_ids.json` — a seeded registry of 17 examine-ID ↔ scene-ID pairs for game objects used in guidance steps, with MCP-verified confidence ratings, sources, and notes.
- Adds `VerifiedSceneId` (Lombok `@Value`) and `VerifiedSceneIdRegistry` loader with `lookupByExamineId`, `lookupBySceneId`, and `all()` methods.
- Adds `VerifiedSceneIdRegistryTest` (19 tests) covering load integrity, lookup by both ID types, unknown-ID returns empty, schema version, immutability, and divergence detection.

## What changed

| File | Type | LOC |
|------|------|-----|
| `src/main/resources/com/collectionloghelper/verified_scene_ids.json` | new | 177 |
| `src/main/java/com/collectionloghelper/data/VerifiedSceneId.java` | new | 48 |
| `src/main/java/com/collectionloghelper/data/VerifiedSceneIdRegistry.java` | new | 180 |
| `src/test/java/com/collectionloghelper/data/VerifiedSceneIdRegistryTest.java` | new | 248 |

### Registry entries

The canonical divergence case is **Shellbane Gryphon cave entrance** (examineId 58441 / sceneId 58439), which is the example documented in `feedback_object_id_gotcha.md`. All other 16 entries have no examine/scene divergence but are recorded because their object types, coordinates, or ID-sourcing history were historically confused or required authoring-log verification.

Full entry list: Shellbane Gryphon cave (58441/58439), Nightmare stairs (32637), Pool of Nightmares (29706), Scorpia cavern (26762), Venenatis cave (47077), CoX entrance (29777), ToB entrance (32653), Corp passage (677), Hallowed Sepulchre door (38574), Duke Sucellus stairs (49100), Leviathan handholds (47592), Whisperer rock (48197), Vardorvis temple entry (48723), Custodian Stalker cave (57279), Fossil Island Wyvern cave (30869), Sarachnis ladder (34862), Sarachnis Thick Web (34858).

### Behavior impact

Purely additive — no existing code paths read from this registry. The loader is not wired into `PluginDataManager` or any production path yet; that is A5.3+ work.

## Test plan

- [ ] `./gradlew test` passes — BUILD SUCCESSFUL, 19 new tests all green
- [ ] `load_succeeds_returnsNonEmptyRegistry` — JSON parses without exception
- [ ] `lookupByExamineId_shellbaneGryphonCacheId_found` — 58441 resolves
- [ ] `lookupBySceneId_shellbaneGryphonSceneId_found` — 58439 resolves
- [ ] `shellbaneEntry_examineIdDiffersFromSceneId` — divergence confirmed in data
- [ ] `lookupByExamineId_unknownId_empty` / `lookupBySceneId_unknownId_empty` — empty Optional
- [ ] `all_returnsUnmodifiableList` — mutation blocked
- [ ] `empty_returnsRegistryWithNoEntries` — factory works
- [ ] All-entries integrity checks pass (objectName, examineId > 0, sceneIds non-empty, confidence, source)

## Milestone reference

Part of A5.2 in docs/ROADMAP.md (Tier A.5). Mark A5.2 done after merge.

### DIFF
--- a/src/main/java/com/collectionloghelper/data/VerifiedSceneId.java
+++ b/src/main/java/com/collectionloghelper/data/VerifiedSceneId.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2025, cha-ndler
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.collectionloghelper.data;
+
+import java.util.List;
+import lombok.Value;
+
+/**
+ * A single entry in the verified scene-ID registry.
+ *
+ * <p>{@code examineId} is the ID reported by the OSRS cache examine tooltip or cache viewer.
+ * {@code sceneIds} are the actual IDs exposed by RuneLite's scene tile API
+ * ({@code Tile.getGameObjects()}, {@code Tile.getWallObject()}, etc.).
+ *
+ * <p>When {@code examineId} differs from {@code sceneIds}, guidance steps and
+ * {@code ObjectHighlightOverlay} must use a value from {@code sceneIds} — not the examine ID.
+ */
+@Value
+public class VerifiedSceneId
+{
+	String objectName;
+	int examineId;
+	List<Integer> sceneIds;
+	String confidence;
+	String action;
+	String canonicalLocation;
+	String source;
+	String notes;
+}
--- a/src/main/java/com/collectionloghelper/data/VerifiedSceneIdRegistry.java
+++ b/src/main/java/com/collectionloghelper/data/VerifiedSceneIdRegistry.java
@@ -1,0 +1,177 @@
+/*
+ * Copyright (c) 2025, cha-ndler
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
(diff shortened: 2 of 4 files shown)


cha-ndler:
## Review — cha-ndler/collection-log-helper#342

**Scope**: JSON schema design · loader error handling · test coverage  
**Result**: 0 CRITICAL · 2 HIGH · 2 MEDIUM · 1 LOW

---

### HIGH — NPE on missing `version` field (loader)

**File**: `VerifiedSceneIdRegistry.java:112`

```java
int version = root.get("version").getAsInt();  // NullPointerException if field absent
```

`root.get("version")` returns `null` when the field is missing from the JSON (e.g. corrupted resource, a future hand-edit that drops the field). `getAsInt()` on null throws `NPE`, which is caught by the outer `catch (Exception e)` and degrades gracefully to `empty()` — so the plugin does not crash. However it emits a misleading error log (`Failed to load verified_scene_ids.json`) instead of a clear schema error, and it silently skips all 17 entries.

Since the registry is currently unused in production paths this is not a runtime risk today, but it will matter when A5.3 wires it in. Recommend a null-guard before the `.getAsInt()` call:

```java
if (!root.has("version")) {
    log.warn("verified_scene_ids.json missing 'version' field; loading anyway");
} else {
    int version = root.get("version").getAsInt();
    if (version != EXPECTED_SCHEMA_VERSION) { ... }
}
```

---

### HIGH — `schemaVersion_isCheckedOnLoad` test does not actually test version checking

**File**: `VerifiedSceneIdRegistryTest.java:203-208`

The test body is:
```java
assertFalse("A successful load must yield entries", registry.all().isEmpty());
```

This is a tautology — it restates `load_succeeds_returnsNonEmptyRegistry` and exercises nothing about version-field handling. The milestone requirement is "JSON schema version field is checked." A real test for this would use a synthetic JSON string with `"version": 99` (wrong version) and confirm the loader still returns entries (current behavior: warn and continue) or that it returns `empty()` (if you ever decide to reject future versions). Without this, there is no regression protection if the version-check path is accidentally deleted.

Suggested fix — add a package-private test helper that parses inline JSON, and write:

```java
@Test
public void load_wrongSchemaVersion_loadsAnywayWithWarning()
{
    // version field present but unexpected — loader must warn and still return entries
    String json = "{\"version\":99,\"entries\":[{\"objectName\":\"Test\",\"examineId\":1,"
        + "\"sceneIds\":[1],\"confidence\":\"low\",\"action\":\"Enter\","
        + "\"canonicalLocation\":\"x\",\"source\":\"test\",\"notes\":\"n\"}]}";
    VerifiedSceneIdRegistry r = VerifiedSceneIdRegistry.load(new Gson(), json);
    assertFalse(r.all().isEmpty());
}
```

This requires a small overload on `load()` that accepts a raw string, or a package-private method that accepts an `InputStream`.

---

### MEDIUM — `confidence` is an unvalidated free-form string in the schema

**File**: `verified_scene_ids.json` (all entries)

All 17 entries use `"confidence": "high"`. The schema has no enumeration of valid values, so a future contributor could write `"confidence": "maybe"` or `"confidence": "verified"` and nothing would catch it. If `confidence` is intended to drive lint tooling behaviour (the stated purpose of this registry), misspelled values would silently produce wrong behaviour.

Consider adding either:
- A `confidence` enum in the JSON schema comment (e.g. note in `description` field: valid values are `high`, `medium`, `low`)
- A Java `enum Confidence { HIGH, MEDIUM, LOW }` in `VerifiedSceneId` and a corresponding Gson deserializer
- Or at minimum a validation step in `VerifiedSceneIdRegistry` that warns on unrecognised confidence values

---

### MEDIUM — `description` field in JSON root is loaded but not stored

**File**: `verified_scene_ids.json:4` / `VerifiedSceneIdRegistry.java`

The JSON root has a `description` field with human-readable context. The loader reads `root.get("entries")` but never reads `root.get("description")`. This is harmless now (Gson ignores unknown fields), but it means the description is not available to A5.3 lint tooling if it needs to surface it in CI output. Consider storing it on the registry or noting explicitly in the class Javadoc that it is intentionally ignored.

---

### LOW — `new JsonParser().parse(reader)` is deprecated in Gson 2.8.6+

**File**: `VerifiedSceneIdRegistry.java:109`

The no-arg constructor of `JsonParser` and its instance `parse()` method were deprecated in Gson 2.8.6 in favour of the static `JsonParser.parseReader()`. The bundled Gson 2.8.5 does not have the static method (hence the current workaround), but if RuneLite bumps to 2.8.6+ the IDE will flag this as deprecated. This is a forward-compatibility concern only, not a bug today. A comment on the line explaining the version constraint would help future maintainers:

```java
// Gson 2.8.5 does not have the static JsonParser.parseReader(); use instance form
JsonElement parsed = new JsonParser().parse(reader);
```

---

### Passing checks

- No personal info, no Claude attribution in commits or code.
- All cross-repo references use the fully-qualified form `cha-ndler/collection-log-helper#NNN`.
- `@Value` + `List<Integer>` on `VerifiedSceneId` is correctly immutable via Lombok.
- `Collections.unmodifiableList` / `Collections.unmodifiableMap` on both indexes is correct.
- `putIfAbsent` on duplicate examine/scene IDs is the right merge strategy.
- `try-with-resources` on the `InputStream` is correct; `InputStreamReader` does not need its own try block since closing the `InputStream` is sufficient here.
- `catch (Exception e)` + `return empty()` is the right degradation strategy for a resource-only registry that must never crash the plugin.
- All 19 tests pass. The per-entry integrity checks (objectName, examineId > 0, sceneIds non-empty, confidence non-blank, source non-blank) provide solid schema regression protection.
- Divergence-detection test (`shellbaneEntry_examineIdDiffersFromSceneId`) is the most important test in the suite and is correctly written.
- No security concerns — file IO is from bundled classpath resource, no user input.

---

**Merge recommendation**: Address the 2 HIGH items before merge. The NPE risk in `root.get("version").getAsInt()` is easy to fix with a null guard; the schema-version test needs a meaningful assertion. MEDIUM items can follow in a subsequent PR.

cha-ndler:
Follow-up commit `471ad722` addresses both HIGH findings from the reviewer pass:

**HIGH #1 — NPE on absent `version` field**: extracted `parseRoot` helper and null-guarded the version element. Missing `version` now logs a warning and assumes schema version 1 instead of NPE'ing into the broad catch and silently returning empty.

**HIGH #2 — tautological schema test**: replaced `schemaVersion_isCheckedOnLoad` with three real tests against a new package-private `loadFromJson(String, Gson)` seam:
- `loadFromJson_unsupportedVersion_stillReturnsEntries` — version=99 accepted with warning
- `loadFromJson_missingVersion_stillReturnsEntries` — regression guard for HIGH #1
- `loadFromJson_malformedJson_returnsEmpty` — exercises the broad-catch fallback

MEDIUM and LOW items left for a follow-up PR so this one stays focused on the HIGH blockers.

`./gradlew test` — BUILD SUCCESSFUL, all tests green.

PR #351, opened 2026-04-17
refactor(panel): A1b — shell widget decomposition (6 widgets, 1288→698 LOC)

cha-ndler:
## Summary

- Extracts 6 shared UI widgets from `CollectionLogHelperPanel` into `com.collectionloghelper.ui.widget`:
  - `SyncStatusView` — sync status label + data-sync warning label
  - `ClueSummaryView` — casket/clue completion summary label
  - `GuidanceBannerView` — guidance active banner + clue guidance banner
  - `StepProgressView` — step progress label, required-items icon strip, Next Step / Skip buttons
  - `SlayerStrategyView` — slayer task indicator + collapsible strategy advisor panel
  - `QuickGuidePanelView` — "Top Pick" quick-guide panel factory with guide-button state tracking
- Removes duplicate `buildControlsPanel()` / `buildContentPanel()` stubs left over from a failed builder-method extraction attempt (Java `final` fields cannot be assigned from non-constructor methods)
- Drops five constructor-only dependency fields from the shell (passed directly to widget constructors)
- Shell LOC: **1288 → 698** (−590 lines, −46%)
- Tests: **468 → 511** (+43 new tests across 6 new test classes, all green)

## Notes on LOC target

The milestone target of <600 LOC is not achievable while preserving `final` fields in the shell constructor. The constructor alone is ~175 lines because Java requires `final` fields to be assigned inline in the constructor body. Reaching <600 would require either accepting non-final mutable fields or a builder/factory pattern that changes the public API. The 698 result is a substantive improvement (−46%) and all real logic is now in focused widget classes.

## Test plan

- [ ] `./gradlew test` passes (511 tests, all green)
- [ ] `./gradlew compileJava` passes with no errors
- [ ] Each new widget has dedicated unit tests covering construction, core state updates, null/empty inputs, and EDT safety
- [ ] `CollectionLogHelperPanel` compiles and delegates correctly to all 6 widgets

### DIFF
--- a/src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java
+++ b/src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java
@@ -51,6 +51,7 @@ import com.collectionloghelper.ui.mode.PetHuntModeController;
 import com.collectionloghelper.ui.mode.SearchModeController;
 import com.collectionloghelper.ui.mode.StatisticsModeController;
 import com.collectionloghelper.ui.widget.ClueSummaryView;
+import com.collectionloghelper.ui.widget.GuidanceBannerView;
 import com.collectionloghelper.ui.widget.SyncStatusView;
 import java.util.EnumMap;
 import java.util.Map;
@@ -160,15 +161,10 @@ public class CollectionLogHelperPanel extends PluginPanel implements PanelShellC
 	private final JPanel listView;
 	private final JPanel detailView;
 
-	private final JPanel clueGuidanceBanner;
-	private final JLabel clueGuidanceLabel;
+	private final GuidanceBannerView guidanceBannerView;
 	private final JPanel slayerStrategyPanel;
 	private final JLabel slayerStrategyLabel;
 	private boolean slayerStrategyExpanded = false;
-
-	private final JLabel guidanceBannerLabel;
-	private final JLabel requirementsWarningLabel;
-	private final JPanel guidanceBannerPanel;
 
 	private final JPanel stepProgressPanel;
 	private final JLabel stepProgressLabel;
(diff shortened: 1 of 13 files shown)


cha-ndler:
## PR Review — A1b Shell Widget Decomposition

**Verdict: APPROVE with one MEDIUM fix and two LOW notes.**

---

### MEDIUM — Unused `Arrays` import in two test files

`SlayerStrategyViewTest.java` and `StepProgressViewTest.java` both import `java.util.Arrays` but never call `Arrays.*` anywhere in the file. `GuidanceBannerViewTest.java` imports it and does use it (`Arrays.asList`), so that one is fine.

```
src/test/java/.../SlayerStrategyViewTest.java:29:  import java.util.Arrays;   ← unused
src/test/java/.../StepProgressViewTest.java:30:   import java.util.Arrays;   ← unused
```

These won't affect compilation or tests but they'll fail a Checkstyle/PMD unused-import rule and are noisy. Remove both before merge.

---

### LOW — `SlayerStrategyViewTest` doesn't cover the expanded code path

The widget has a non-trivial expanded state (`expanded = true` → `updateStrategyContent` builds a full HTML string with sources, block list, etc.). All four tests leave `expanded = false`. The collapsed path is tested; the expanded path is not. Not a merge blocker since the widget is a direct extraction of existing code, but a follow-up test covering the expanded branch would be good hygiene.

---

### LOW — `QuickGuidePanelView` button listener doesn't notify shell of state change

In the OLD code the guide button's action listener called `setGuidanceState(false/true, ...)` directly on the shell, which kept `guidanceActive` / `guidedSource` in sync immediately. In the NEW code the listener only calls `guidanceDeactivator.run()` or `guidanceActivator.accept(source)`, which routes through `GuidanceOverlayCoordinator` which then calls `panel.setGuidanceState()`. This is architecturally more correct (coordinator owns the state), but it introduces a round-trip: the button updates its own text immediately, then `syncGuidanceState` may re-sync from an async coordinator callback. In practice this should be fine because the coordinator's `setGuidanceState` path uses `invokeLater`, but the behavior differs subtly from the original. Worth a comment in `QuickGuidePanelView.create()` noting that the button's visual state is maintained both locally (immediate) and externally (via `syncGuidanceState`), so double-update is expected and harmless.

---

### Passing checks

- **Behavior preservation**: all 6 widgets are direct extractions — no logic changes, color constants preserved verbatim
- **Thread safety**: all public mutators on widgets that can be called off-EDT use `SwingUtilities.invokeLater`; `QuickGuidePanelView.create()` correctly documents EDT requirement
- **Snapshot discipline**: `SyncStatusView.updateDataSyncWarning()` snapshots `isFullySynced`, `isCollectionLogSynced`, `isBankScanned` before dispatch — correct
- **Null safety**: `ClueSummaryView.updateFromBankState(null)` handled; `GuidanceBannerView.showGuidance(null)` handled via early exit; `QuickGuidePanelView.syncGuidanceState` null-guards `lastGuideButton` and `lastSource`
- **`final` field constraint**: correctly documented in PR body — not achievable to go below 600 LOC without dropping `final` fields or using a builder
- **No personal info**: copyright header uses `cha-ndler` alias only — clean
- **No Claude attribution**: commit messages and source files are clean
- **LOC delta**: 1288 → 698 (−46%) is a legitimate improvement; the remaining bulk is the constructor's `final` field initialization which is a Java language constraint
- **Test count**: 468 → 511 (+43), all new widget test classes follow the established "doesNotThrow + edge case" pattern used by the rest of the test suite
- **Security**: no user input handling, no network calls, no file IO — nothing to flag

cha-ndler:
Follow-up commit `5abb8a5b` addresses the MEDIUM finding:

Removed unused `java.util.Arrays` imports from `SlayerStrategyViewTest.java` and `StepProgressViewTest.java`. `GuidanceBannerViewTest.java` genuinely uses `Arrays.` so its import stays.

`./gradlew test` — BUILD SUCCESSFUL, 511 tests green.

LOW items (`SlayerStrategyView` expanded-state test coverage + `QuickGuidePanelView` double-update interaction comment) left for a follow-up PR so this one stays focused on the MEDIUM blocker.

PR #354, opened 2026-04-18
fix(panel-widget): rename StepProgressView.hide() to hideStep() (#353)

### LINKED ISSUE #353: Step details widget visible when no guidance is active (logged out, idle, after Stop)
## Summary

The step details widget (containing the **Skip** button and step text) is visible in panel states where it should be hidden. It appears pre-login, when logged in with no guidance running, and persists stale content after guidance is stopped.

Likely a regression from the A1b shell widget decomposition (#351) — visibility wiring for the step-details widget probably wasn't preserved when splitting the shell.

## Repro

1. **Logged out** — open RuneLite, open the Collection Log Helper side panel without logging in.
   - **Expected:** step details widget hidden (no Skip button).
   - **Actual:** blue-outlined box with a **Skip** button is visible under the sync header.

2. **Logged in, no guidance active** — log into the game; do not start guidance.
   - **Expected:** step details widget hidden.
   - **Actual:** empty blue box with **Skip** button shows below the purple Slayer Strategy box.

3. **Guidance active** (control case — works correctly) — click **Guide Me**.
   - Widget populates with ``Guiding: <target>``, ``Step N/M: ...``, and Next Step / Skip buttons as expected.

4. **Guidance stopped** — click **Stop Guidance**.
   - **Expected:** step details widget hidden and cleared.
   - **Actual:** widget stays visible and still shows the last step text (e.g. ``Step 3/5: Pick up your shade key from the reward pillar``) with Next Step / Skip buttons, even though the ``Guiding:`` header is gone and the green **Guide Me** button has returned.

## Context

- Panel mode: default (Slayer Strategy surface visible).
- Character: reproduces both logged out and logged in (Hanadji, in-game screenshots).
- Version: current master including #351 (A1b shell widget decomposition, merged 2026-04-17).

## Suspected area

A1b decomposed the panel shell into 6 widgets. The step-details widget's ``setVisible`` / hide path is likely not being called on:
- initial panel build before any guidance exists,
- guidance-stopped transition.

Check the step-details widget's visibility wiring against the pre-A1b shell (commit 3b4c71f9 vs 9b62bd7b) — the old shell either gated visibility on an ``activeGuidance != null`` check or only added the component to the layout on start.

## Screenshots

_Attached in follow-up comment — four states: logged out, logged in idle, guidance active (correct), guidance stopped (bug persists)._

cha-ndler:
## Summary

Fixes #353. The step-details widget (blue-bordered panel with the **Skip** button) was visible in every state where guidance wasn't running — logged out, logged in idle, and after Stop Guidance still showing stale step text.

**Root cause** (not what the issue body guessed): `StepProgressView` was extracted in #351 with a `public void hide()` method. That signature shadows the deprecated-but-still-wired `java.awt.Component#hide()`. `Component.setVisible(boolean)` is implemented as `show(b)` which calls `hide()` when `b == false`, so Java's virtual dispatch routed it to the widget's override — which only schedules an `invokeLater` and never flips the `visible` field. Every call to `setVisible(false)` on this widget — including the one in its own constructor — was a silent no-op.

Why it worked pre-A1b: the old `stepProgressPanel` was a raw `JPanel`, not a subclass, so nothing shadowed `Component.hide()`.

Why existing tests missed it: `StepProgressViewTest` only asserted `doesNotThrow`, never `isVisible()`.

**Fix:** rename `hide()` → `hideStep()`, update the one caller (`CollectionLogHelperPanel.hideStepProgress`), and add three visibility regression tests that flush the EDT and assert `isVisible()`.

## Test plan

- [x] New test `afterConstruction_isNotVisible` — widget starts hidden
- [x] New test `hideStep_afterShowStep_actuallyHides` — `isVisible() == false` after `hideStep()`
- [x] New test `showStep_makesVisible` — `isVisible() == true` after `showStep(...)`
- [x] Manual smoke: pre-login, logged-in idle, after Stop Guidance — Skip panel gone in all three states
- [x] Manual smoke: Guide Me → panel populates → Next Step / Skip buttons still wired and functional

## Notes for reviewer

The method is renamed rather than refactoring the widget to follow the outer-transparent / inner-panel pattern used by `GuidanceBannerView` et al. That refactor would also fix it but touches more surface area; the rename addresses the actual root cause with minimal diff. Class-level Javadoc calls out the shadowing pitfall so it doesn't recur.

### DIFF
--- a/src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java
+++ b/src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java
@@ -627,7 +627,7 @@ public class CollectionLogHelperPanel extends PluginPanel implements PanelShellC
 	/** Hides the step progress banner. */
 	public void hideStepProgress()
 	{
-		stepProgressView.hide();
+		stepProgressView.hideStep();
 	}
 
 	/** Sets callbacks for step advance/skip buttons. */
--- a/src/main/java/com/collectionloghelper/ui/widget/StepProgressView.java
+++ b/src/main/java/com/collectionloghelper/ui/widget/StepProgressView.java
@@ -52,8 +52,14 @@ import net.runelite.client.util.AsyncBufferedImage;
  * and the Next Step / Skip buttons.
  *
  * <p>Constructed once by the shell; updated via {@link #showStep} and
- * hidden via {@link #hide}. Step-advance and skip callbacks are wired
+ * hidden via {@link #hideStep}. Step-advance and skip callbacks are wired
  * via {@link #setCallbacks}.
+ *
+ * <p><b>Note:</b> the hide method is named {@code hideStep} rather than
+ * {@code hide} to avoid shadowing the deprecated {@link java.awt.Component#hide()}.
+ * {@code Component.setVisible(false)} internally dispatches to {@code hide()};
+ * an override with the same signature breaks {@code setVisible(false)} for this
+ * widget and leaves it visible in pre-guidance states (see issue #353).
  */
 public class StepProgressView extends JPanel
 {
@@ -185,8 +191,12 @@ public class StepProgressView extends JPanel
 	/**
 	 * Hides the step progress panel and clears required items.
 	 * Safe to call from any thread.
+	 *
+	 * <p>Intentionally named {@code hideStep} (not {@code hide}) — see class
+	 * Javadoc and issue #353 for why overriding {@link java.awt.Component#hide()}
+	 * would break {@link #setVisible}.
 	 */
-	public void hide()
+	public void hideStep()
 	{
 		SwingUtilities.invokeLater(() ->
 		{
--- a/src/test/java/com/collectionloghelper/ui/widget/StepProgressViewTest.java
+++ b/src/test/java/com/collectionloghelper/ui/widget/StepProgressViewTest.java
@@ -28,11 +28,14 @@ import com.collectionloghelper.data.PlayerBankState;
 import com.collectionloghelper.data.PlayerInventoryState;
 import net.runelite.client.game.ItemManager;
 import java.util.Collections;
+import javax.swing.SwingUtilities;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mockito;
 
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
 /**
  * Unit tests for {@link StepProgressView}.
@@ -78,9 +81,9 @@ public class StepProgressViewTest
 	}
 
 	@Test
-	public void hide_doesNotThrow()
+	public void hideStep_doesNotThrow()
 	{
-		view.hide();
+		view.hideStep();
 	}
 
 	@Test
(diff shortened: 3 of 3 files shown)


cha-ndler:
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#354
**Files changed**: 3 (+60/-7)
**Reviewer focus**: Verify the rename addresses the shadowing bug end-to-end, confirm no other widget has the same pitfall, score the new `isVisible()` regression tests, and check privacy/attribution hygiene.

### Findings

#### CRITICAL (0)
- None.

#### HIGH (0)
- None.

#### MEDIUM (0)
- None.

#### LOW (2)
- **Tiny Javadoc nit** (optional): in `StepProgressView.java` line 197, `{@link #setVisible}` resolves unambiguously today because only `Component.setVisible(boolean)` is in scope, but `{@link #setVisible(boolean)}` would be more explicit and future-proof against any future overload.
- **Pre-existing issue unrelated to this PR**: `./gradlew javadoc` fails with one error in `ClueSummaryView.java:42` (reference not found). Not introduced by this PR — flagging so it can be cleaned up separately.

### Plugin-specific patterns
- No `drop_rates.json` / guidance / NPC / object-ID / efficiency changes. Validators not applicable.
- Rename preserves behaviour for the happy path: `showStep`, `setCallbacks`, and button-click dispatch are untouched. Only the misnamed method and its single caller change.
- Confirmed no other widget in `src/main/java/com/collectionloghelper/ui/widget/` defines `public void hide()` — grep returned zero matches, so this bug is unique to `StepProgressView`.
- All `.hide()` callsites in `src/` are gone — grep for `\.hide\(\)` returned no matches.
- Only `stepProgressView.hideStep()` is called now (line 630 of `CollectionLogHelperPanel.java`); no stale caller left behind.

### Verification ran
- [x] `./gradlew test` — 514/514 passed, 0 failures, 0 ignored
- [x] `StepProgressViewTest` — 11/11 passed (includes 3 new regression tests: `afterConstruction_isNotVisible`, `hideStep_afterShowStep_actuallyHides`, `showStep_makesVisible`)
- [x] Grep for `public void hide()` in `src/main/java` — zero matches (no other widget shadows `Component#hide()`)
- [x] Grep for `\.hide\(\)` in `src/` — zero matches (no stale callers)
- [x] Grep for `stepProgressView\.` — only `setAlignmentX`, `showStep`, `hideStep`, `setCallbacks` — all expected
- [x] Commit authorship — `cha-ndler <48898494+cha-ndler@users.noreply.github.com>` across all 5 recent commits; no personal info, no Claude attribution
- [x] Grep for "Chandler" / "ch@ndler.net" in `src/` — zero matches
- [x] Cross-repo refs — PR body references `#353` and `#351`, both same-repo and correct
- [x] Javadoc for modified file — `{@link java.awt.Component#hide()}` and `{@link #setVisible}` both resolve cleanly (no warnings/errors on `StepProgressView.java`)

### Test quality assessment
The three new regression tests are exactly the right shape:

1. **`afterConstruction_isNotVisible`** — catches the original bug directly. The constructor at line 98 calls `setVisible(false)`; this test pins that contract. Would have caught this regression the moment it shipped.
2. **`hideStep_afterShowStep_actuallyHides`** — the canonical round-trip: show then hide, asserting both sides. This is the test that would have failed in the original bug state (visible=true after hide).
3. **`showStep_makesVisible`** — complementary positive-path assertion to confirm `showStep` flips the flag.

**EDT flush via `SwingUtilities.invokeAndWait(() -> {})`** is the correct primitive here. Both `showStep` and `hideStep` wrap their work in `invokeLater`, and `invokeAndWait` blocks until prior EDT tasks drain, so by the time the assertion runs the `setVisible` call has completed. On headless CI (`GraphicsEnvironment.isHeadless() == true`), `JPanel` construction and `setVisible` both still work — they just don't get a peer — so this should not flake. Confirmed by 11/11 tests passing locally on Windows.

One minor consideration (not a blocker): if `showStep`/`hideStep` ever get refactored to chain through a `CompletableFuture` or a second `invokeLater`, the single `invokeAndWait` would no longer be sufficient. Leaving the `flushEdt()` helper as a named method is the right call — future maintainers can expand it in one place if needed.

### Recommendation
**SHIP** — no CRITICAL/HIGH/MEDIUM issues. The rename is minimal, the root cause diagnosis matches the empirical evidence, the Javadoc explains the pitfall clearly enough that future maintainers will not reintroduce it, and the regression tests would have caught the original bug. Two trivial LOW-severity notes; neither blocks merge.

### Notes
- Strongly consider adopting this pattern elsewhere: **any class extending a `java.awt.Component` should never define methods named `hide()`, `show()`, `size()`, `resize()`, `enable()`, `disable()`, `reshape()`, `move()`, `minimumSize()`, `preferredSize()`, `maximumSize()`, `bounds()`, or `layout()`** — all of these are deprecated-but-still-dispatched Component methods. A lightweight checkstyle rule forbidding overrides with those names (or a unit test that greps for them in widget files) would prevent recurrence. Worth adding to `feedback_duplicate_overlay_init.md` or a new `feedback_awt_component_name_shadowing.md` memory file.
- The class-level Javadoc note is well-written and explains the *why* (not the *what*), matching the project's comment-quality standard.
- The commit message is excellent — it tells the story, names the mechanism, and explains why existing tests missed it. Consider this as a template for future root-cause commits.

PR #359, opened 2026-04-18
feat(guidance): B1 — composable AND/OR completion conditions (additive)

cha-ndler:
Closes B1 in docs/ROADMAP.md.

## Summary

Introduces a composable AND/OR completion tree for guidance steps. The existing 10-value `CompletionCondition` enum becomes the atomic leaf (`CompletionNode.Leaf`) of a new discriminated union (`Leaf | All | Any`). Every production source in `drop_rates.json` stays on the legacy single-enum shape; the new tree is strictly additive and opt-in per step via a new `completionNode` field.

## Design

- **Java 11 + Lombok** — no sealed interfaces. Used an abstract `CompletionNode` class with package-local constructor and three final subtypes (`Leaf`, `All`, `Any`) to get a closed hierarchy.
- **Gson adapter** (`CompletionNodeAdapter`) accepts three input shapes:
  1. Legacy bare string → coerced to `Leaf(CompletionCondition)`. This is what every production source uses today; adapter wiring is transparent.
  2. `{ \"op\": \"leaf\", \"condition\": \"...\" }` — explicit leaf form.
  3. `{ \"op\": \"all\"|\"any\", \"nodes\": [...] }` — branching form with recursive parse. Accepts `children` as an author-friendly alias for `nodes`.
- **Evaluator** — `GuidanceSequencer.isStepAlreadySatisfied()` now checks `step.getCompletionNode()` first. When present, it walks the tree via a shared `evaluateLeafCondition(step, condition)` predicate. When absent (every production step today), the existing legacy switch runs unchanged. Event-triggered handlers (item drops, NPC death, chat, varbit) are untouched and continue to drive through `step.getCompletionCondition()`.

## Backward compatibility

- **Zero data migration.** The 225 production sources stay on the legacy shape.
- `CompletionNodeRegressionTest` loads the real `drop_rates.json` twice through independent `DropRateDatabase` instances and asserts the `(sourceName, stepIndex) -> completionCondition.name()` snapshot is identical both times, AND that `completionNode == null` for every step. This is the pre/post behavioral proof the B1 spec asked for.
- `validate_drop_rates` and `guidance_lint` both report no new issues.

## Pilot fixture

`src/test/resources/com/collectionloghelper/data/b1_pilot_source.json` is the single pilot source exercising the new shape end to end. It includes:
- a legacy-shape step (string enum, no tree),
- an AND step requiring inventory AND zone arrival,
- an OR step accepting either item obtained OR inventory has,
- a nested ALL(ITEM_OBTAINED, ANY(INVENTORY_HAS_ITEM, PLAYER_ON_PLANE)).

## Test plan

- [x] `./gradlew test` — 563/563 passing (baseline was 516; +47 new).
  - `CompletionNodeTest` — 16 tests covering Leaf/All/Any evaluation, short-circuit, empty semantics, defensive copies, equals/hashCode.
  - `CompletionNodeAdapterTest` — 15 tests covering legacy string, explicit leaf, branching, nested, round-trip, error cases.
  - `GuidanceSequencerCompletionNodeTest` — 7 tests proving tree-driven skip-ahead, tree precedence over legacy enum, and legacy-path preservation.
  - `CompletionNodeRegressionTest` — 3 tests asserting production data is additive-only + snapshot-stable.
  - `CompletionNodePilotFixtureTest` — 6 tests parsing the pilot fixture end to end and evaluating its trees.
- [x] `validate_drop_rates` — only pre-existing duplicate-itemId warnings, nothing new.
- [x] `guidance_lint --severity critical` — 0 issues across 225 sources.

## Commits

- `feat(data): add CompletionNode AND/OR tree types for B1 (additive)` — `92633430`
- `feat(guidance): evaluate CompletionNode trees in sequencer (B1)` — `a94cc6ad`
- `test(guidance): CompletionNode unit + regression + pilot fixture (B1)` — `a8ac9285`
- `docs: mark B1 in-progress in status log` — `cd56886a`

## Follow-up (out of scope)

- Migrating any production source to the new shape is a separate tier-D sweep.
- The evaluator still returns `false` for event-only leaves (`ACTOR_DEATH`, `CHAT_MESSAGE_RECEIVED`, `VARBIT_AT_LEAST`, `NPC_TALKED_TO`, `MANUAL`) when polled eagerly — matching the legacy switch exactly. Teaching the tree path to latch event-driven leaves would be a separate change and is not required by B1.

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -511,7 +511,7 @@ Focus: god-object decomposition, build hardening, test expansion. Resulted in th
 
 **Tier B — Guidance depth parity with Quest Helper**
 
-- [ ] B1 — Composable completion conditions             status: planned       owner: —          updated: 2026-04-16
+- [/] B1 — Composable completion conditions             status: in-progress   owner: cha-ndler  updated: 2026-04-17
 - [ ] B2 — Tile-sequence pathing                        status: planned       owner: —          updated: 2026-04-16
 - [ ] B3 — Nested conditional steps                     status: planned       owner: —          updated: 2026-04-16
 - [ ] B4 — Alternative-method modelling                 status: planned       owner: —          updated: 2026-04-16
--- a/src/main/java/com/collectionloghelper/data/CompletionNode.java
+++ b/src/main/java/com/collectionloghelper/data/CompletionNode.java
@@ -1,0 +1,264 @@
+/*
+ * Copyright (c) 2025, cha-ndler
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.collectionloghelper.data;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Predicate;
+
+/**
+ * A composable tree node describing when a guidance step should be considered complete.
+ *
+ * <p>This is the tagged union that powers B1 ("Composable AND/OR completion conditions").
+ * Java 11 does not support {@code sealed} interfaces, so the closed hierarchy is expressed
+ * via an abstract class with a finite set of package-local subtypes:
+ *
+ * <ul>
+ *   <li>{@link Leaf} — wraps a single {@link CompletionCondition} atomic check.</li>
+ *   <li>{@link All} — true when every child node evaluates true.</li>
+ *   <li>{@link Any} — true when any child node evaluates true.</li>
+ * </ul>
+ *
+ * <p>The existing single-enum {@code completionCondition} field on {@link GuidanceStep}
+ * remains the canonical legacy shape. When a step defines a {@link CompletionNode}, the
+ * sequencer walks the tree for eager (polled) completion checks; leaf evaluation matches
+ * the semantics of the legacy switch exactly, so a step written as the legacy single enum
+ * or as a {@code Leaf} of that enum behaves identically.
+ *
+ * <p>Instances are immutable. {@link All} and {@link Any} defensively copy and wrap their
+ * children with {@link Collections#unmodifiableList(List)} on construction.
+ */
+public abstract class CompletionNode
+{
+	/** Node kind tag — useful for logging, equality, and discriminated serialization. */
+	public enum Kind
+	{
+		LEAF,
+		ALL,
+		ANY
+	}
+
+	CompletionNode()
+	{
+		// Package-private constructor: only the nested subtypes may extend.
+	}
+
+	public abstract Kind getKind();
+
+	/**
+	 * Walks the node tree, evaluating every leaf via the supplied predicate.
+	 *
+	 * @param leafEvaluator maps a {@link CompletionCondition} to a boolean pass/fail
+	 * @return {@code true} when the composed tree evaluates to complete
+	 */
+	public abstract boolean evaluate(Predicate<CompletionCondition> leafEvaluator);
+
+	public static Leaf leaf(CompletionCondition condition)
+	{
+		return new Leaf(condition);
+	}
+
+	public static All all(List<CompletionNode> children)
+	{
+		return new All(children);
+	}
+
+	public static Any any(List<CompletionNode> children)
+	{
+		return new Any(children);
+	}
+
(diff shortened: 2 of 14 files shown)


cha-ndler:
**Holding for v1.0.0-hub tag push.**

This PR implements B1 (composable AND/OR completion conditions, additive schema) and is functionally complete — 563/563 tests passing, additive-only, 225 production sources unchanged. Self-review on the worker side flagged 0 CRITICAL / 0 HIGH / 0 MEDIUM / 3 LOW nits.

Per the pipeline gating rule, Tier B and D work does not merge until the `v1.0.0-hub` tag is pushed to `origin` (verified via `git ls-remote --tags origin v1.0.0-hub`). The tag currently exists only locally; the remote does not have it.

The branch stays alive. No rework is expected once the gate opens — this PR just flips from draft to ready and picks up the standard `clh-pr-reviewer` pass at that point.

Reference: `pipeline/WORKER_TEMPLATE.md` hard-rule block on the Tier B/D gate.

cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#359
**Files changed**: 14  (+1670/-20)
**Reviewer focus**: backward compatibility of Gson adapter, null-safety of the legacy fallback path, tree evaluation correctness, regression test robustness against vacuous-pass.

### Findings

#### CRITICAL (0)

None.

#### HIGH (0)

None.

#### MEDIUM (1)

1. **No recursion depth guard in `CompletionNodeAdapter.readChildren()`** -- The adapter recurses via `ctx.deserialize(child, CompletionNode.class)` with no depth limit. A malformed or malicious JSON file with deeply nested all/any trees could cause a StackOverflowError. In practice the only JSON sources are shipped with the plugin or authored by the developer, so exploitation risk is near zero, but a depth counter (e.g., max 16 levels, threaded via a field on the adapter) would be cheap defensive hardening. Consider adding in a follow-up.

#### LOW (3)

1. **`readOp()` only checks `"op"` and `"OP"`, not mixed-case** -- The adapter normalizes the *value* via `toLowerCase(Locale.ROOT)` but only looks for the key as exactly `"op"` or `"OP"`. Mixed-case keys like `"Op"` or `"oP"` would fall through to the no-op path and either hit the implicit-leaf fallback (if `condition` is present) or throw. The Javadoc says "lenient to field casing" which slightly overpromises. Since the authoring convention is lowercase and this is internal data, the practical risk is nil. Consider updating the Javadoc to say "accepts `op` or `OP`" rather than claiming general casing leniency.

2. **Short-circuit tests use anonymous `CompletionNode` subclass from same package** -- `CompletionNodeTest.all_shortCircuitsOnFirstFalse()` and `any_shortCircuitsOnFirstTrue()` create anonymous subclasses of `CompletionNode` by reaching the package-private constructor. This is fine for testing short-circuit behavior, but it does slightly undermine the "closed hierarchy" contract. A comment acknowledging this is test-only would improve clarity.

3. **Copyright headers say 2025** -- All new files have `Copyright (c) 2025, cha-ndler`. It is 2026. Minor, but worth a batch fix at some point.

### Verification ran

- [x] `./gradlew test` -- 563/563 passed (BUILD SUCCESSFUL)
- [x] No personal info detected (grep for first name and personal email in diff and commit messages)
- [x] No Claude attribution (grep for "co-authored-by", "generated by Claude" in diff and commits)
- [x] Cross-repo refs use fully-qualified form (no bare `#NNN` references found)
- [x] Conventional commit prefixes on all 4 commits (feat, feat, test, docs)
- [x] PR description has summary, design rationale, test plan with checked items

### Specific review-focus answers

**1. Backward compatibility (Gson adapter + legacy coercion)**
Correct. The adapter returns `null` for `JsonNull`, coerces bare strings to `Leaf` via `parseCondition()` which normalizes case. Gson field-level reflection on `@Value GuidanceStep` means the new `completionNode` field defaults to `null` when absent from JSON -- confirmed by both the regression test and pilot fixture test (`legacyStepKeepsNullCompletionNode`). All 225 production sources parse unchanged.

**2. Null safety (completionNode == null fallback)**
Correct. `GuidanceSequencer.isStepAlreadySatisfied()` checks `step.getCompletionNode() != null` before tree evaluation, then falls through to the legacy `step.getCompletionCondition()` switch. The extracted `evaluateLeafCondition()` method also returns `false` for `null` condition. Both paths are exercised by `legacyPathStillWorksWhenNodeAbsent` and the regression test.

**3. Tree evaluation (short-circuit + freeze)**
Correct. `All.evaluate()` returns `false` on first failing child (short-circuit AND). `Any.evaluate()` returns `true` on first passing child (short-circuit OR). Empty `All` is vacuously true, empty `Any` is vacuously false -- both tested. `freeze()` defensively copies to a new `ArrayList` and wraps with `Collections.unmodifiableList()`, with null-entry rejection. Tests cover defensive copy, unmodifiable access, and null entry rejection.

**4. Regression test quality (vacuous-pass risk)**
Not vacuous. `legacyEnumSnapshotPreserved()` builds a `List<String>` of `"sourceName#index=CONDITION"` pairs across *all* parsed sources, then asserts `preSnapshot.size() >= 200` as a sanity floor. With 225 sources and many multi-step guidance sequences, the snapshot covers hundreds of meaningful pairs. The `everyStepCompletionNodeIsNull_additiveOnly` test iterates every step and fails on the first non-null completionNode. This directly guards against accidental migration -- the PR #356 lesson about empty-list vacuous-pass is addressed.

**5. Gson adapter robustness**
Good coverage: unknown condition string throws `JsonParseException`, unknown op throws, missing `nodes` array on branching op throws, object without `op` or `condition` throws, bare numeric primitive throws. Null JSON returns null. Only gap is the recursion depth noted in MEDIUM-1.

### Roadmap alignment

- PR body says "Closes B1 in docs/ROADMAP.md" but the diff only flips B1 to `[/]` (in-progress), not `[x]` (done). This is correct -- the PR is in draft, gated on v1.0.0-hub tag push. No scope creep beyond B1.

### Recommendation

**SHIP** -- no CRITICAL or HIGH issues. The one MEDIUM (recursion depth guard) is a hardening suggestion for a follow-up, not a merge blocker given that all JSON is developer-authored. The three LOW items are cosmetic.

### Notes

- The `evaluateLeafCondition()` extraction from the existing switch is a clean refactor. Adding `NPC_TALKED_TO` and `MANUAL` to the explicit fall-through cases (lines 704-705 in the diff) is a good correctness improvement -- these were previously handled by the `default` case but are now explicit.
- The pilot fixture is a well-designed test artifact that exercises all four step shapes (legacy, AND, OR, nested) in a single source file.
- The `@Value` + Gson combination works because Gson uses `Unsafe`-based field injection, bypassing the constructor. This is a known pattern in the codebase and is safe here since the new field is nullable and last in field order.
- When production sources eventually migrate to tree shapes (tier-D sweep), the `everyStepCompletionNodeIsNull_additiveOnly` test will need to be updated or removed. Consider adding a TODO comment in the test noting this expectation.

cha-ndler:
Closing as stale. B1 is still tracked in [docs/ROADMAP.md (Tier B status log)](https://github.com/cha-ndler/collection-log-helper/blob/master/docs/ROADMAP.md#L548) as `planned` / un-owned, but B2, B3, B5, and B6 shipped without it (PRs #455, #456, #473, #474), and 16 commits have landed on master that touch the same files — including four new `GuidanceStep` fields (`perItemRequiredItemIds`, `perItemStepDescription`, `perItemNpcId`, `guidanceHelperKey`), the B5 dynamic-step dispatch, and the B6 helper-key routing. Rebasing this draft would require re-grafting the sequencer hook onto that newer dispatch surface and updating every positional `GuidanceStep` constructor in the test suite.

The design rationale (Java 11 abstract-class pattern over sealed interfaces, three-shape Gson adapter, additive regression test snapshotting the 225 production sources) is preserved in this PR description for whoever picks B1 up next.

Branch `tier-b/b1-composable-completion-conditions` will be deleted.

Issue #433, opened 2026-05-14
bug: In-game guidance overlay renders em-dash as 'â€"' mojibake (charset issue at draw time)


cha-ndler:
The in-game guidance overlay (top-left in 3D scene viewport) renders em-dash characters as the UTF-8-as-Windows-1252 mojibake sequence 'â€"' instead of the correct '—'.

Confirmed the data file is correct: drop_rates.json's em-dash bytes are e2 80 94 (correct UTF-8). The mojibake is introduced at render time by the overlay text-drawing path.

Reproducer:
1. Activate guidance on Perilous Moons (or any source whose step description contains an em-dash). Mort'ton's step 1 description currently has '—' too.
2. Look at the in-game overlay text.
3. Observe: every em-dash renders as 'â€"' instead of '—'.

Hypothesis: the overlay's text-drawing site reads the description as UTF-8 bytes but renders them via a draw call that treats input as Windows-1252 (or vice-versa — UTF-16 string -> bytes -> wrong-charset render). Likely in GuidanceOverlay's render method where setLocationDescription's value is drawn via Graphics2D.drawString.

Workaround for now: per-item description overrides (PR #421) typically don't include em-dashes, so target-aware activations sidestep the issue. But fallback (static) display still shows the corruption.

Severity: LOW (cosmetic). Surfaced 2026-05-13 in-client validation pass (F7).

PR #507, opened 2026-05-17
ci(lint): fail build on non-ASCII characters in drop_rates.json + replace em-dashes with ASCII (#501)

### LINKED ISSUE #501: ci(lint): fail build on non-ASCII characters in drop_rates.json to prevent em-dash mojibake recurrence
## Problem

The em-dash mojibake incident took **two** PRs to clean up:

- `#433` fix(charset): force UTF-8 when loading JSON resources to prevent em-dash mojibake
- `#443` (same)

The charset fix is correct at the loader layer, but the **content layer** still has no guard. The next fancy character that sneaks into `drop_rates.json` (smart quote, en-dash, ellipsis, non-breaking space) will reopen the same class of bug.

## Proposed fix

Add a CI lint that fails the build if `src/main/resources/com/collectionloghelper/drop_rates.json` contains any non-ASCII byte.

Implementation: add a Gradle task `lintDropRates` wired to `check`, plus an explicit CI step. The check itself is one line:

```bash
! LC_ALL=C grep -P '[^\x00-\x7f]' src/main/resources/com/collectionloghelper/drop_rates.json
```

(Equivalent Groovy/Java check inside Gradle is fine — whichever is more idiomatic for this build.)

## Files

- `.github/workflows/ci.yml` (or `.github/workflows/build.yml`) — add lint step
- `build.gradle` — add `lintDropRates` task; hook into `check`

## Acceptance

- [ ] `./gradlew lintDropRates` passes on clean master
- [ ] Introducing any non-ASCII character into `drop_rates.json` causes `./gradlew lintDropRates` to fail with a clear message naming the file
- [ ] Lint runs on CI for every PR
- [ ] No regression in build time (>1s extra is fine; multi-second slowdown is not)

## Notes

- Other resources are exempt for now — only `drop_rates.json` is in scope. If the lint should expand to other JSON resources, that's a follow-up.
- The "right answer" long-term is to **never emit** non-ASCII content from generation; this lint is the safety net.

cha-ndler:
## Summary

- Replaces 20x U+2014 em-dashes with ASCII `-` in step descriptions in `src/main/resources/com/collectionloghelper/drop_rates.json` (18 lines touched).
- Adds a `lintDropRates` Gradle task that scans `drop_rates.json` for any codepoint > U+007F and fails with line/column/codepoint info. Wired into the `check` lifecycle so `./gradlew build` gates on it.
- Updates `DropRateDatabaseTest.load_guidanceStepDescriptions_noCharsetCorruption` to enforce the new ASCII-only invariant on the loaded model (descriptions + perItemStepDescription overrides), while keeping the cp1252 mojibake guard from issue #433.

Policy: drop_rates.json is ASCII-only. Fancy punctuation (em-dash, smart quotes, ellipsis, non-breaking space) must be replaced with ASCII equivalents.

## Test plan

- [x] `./gradlew lintDropRates` passes on clean content
- [x] Injecting one em-dash into a step description causes `./gradlew lintDropRates` to fail with line/col + codepoint U+2014 in the error message (verified locally, then reverted)
- [x] `./gradlew build` passes end-to-end (lintDropRates + jacocoTestCoverageVerification both run inside `check`)
- [x] Spot-checked 3 of the 18 modified descriptions for natural readability after `-` replacement (no doubled spaces, no broken word breaks)

Closes #501

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -115,6 +115,34 @@ test {
 
 check.dependsOn jacocoTestCoverageVerification
 
+tasks.register('lintDropRates') {
+	group = 'verification'
+	description = 'Fails if drop_rates.json contains non-ASCII characters.'
+	def jsonFile = file('src/main/resources/com/collectionloghelper/drop_rates.json')
+	inputs.file(jsonFile)
+	doLast {
+		if (!jsonFile.exists()) {
+			throw new GradleException("Missing ${jsonFile}")
+		}
+		def violations = []
+		int lineNo = 0
+		jsonFile.eachLine('UTF-8') { line ->
+			lineNo++
+			for (int i = 0; i < line.length(); i++) {
+				int cp = line.codePointAt(i)
+				if (cp > 0x7f) {
+					violations << "  line ${lineNo} col ${i + 1}: codepoint U+${String.format('%04X', cp)} ('${new String(Character.toChars(cp))}')"
+					break
+				}
+			}
+		}
+		if (!violations.isEmpty()) {
+			throw new GradleException("Non-ASCII characters in drop_rates.json:\n" + violations.take(20).join("\n") + (violations.size() > 20 ? "\n  ... and ${violations.size() - 20} more" : "") + "\nPolicy: drop_rates.json must be ASCII-only. Replace fancy punctuation (em-dash, smart quotes, ellipsis, non-breaking space) with ASCII equivalents.")
+		}
+	}
+}
+check.dependsOn lintDropRates
+
 tasks.register('shadowJar', Jar) {
 	dependsOn configurations.runtimeClasspath
 	manifest {
--- a/src/main/resources/com/collectionloghelper/drop_rates.json
+++ b/src/main/resources/com/collectionloghelper/drop_rates.json
@@ -1139,7 +1139,7 @@
         "section": "Travel"
       },
       {
-        "description": "Kill the Alchemical Hydra. Use the coloured vents to weaken each phase — walk over the correct vent matching the Hydra's colour",
+        "description": "Kill the Alchemical Hydra. Use the coloured vents to weaken each phase - walk over the correct vent matching the Hydra's colour",
         "worldX": 1364,
         "worldY": 10265,
         "worldPlane": 0,
@@ -3432,7 +3432,7 @@
         "completionDistance": 20
       },
       {
-        "description": "Kill Dagannoth Rex. Use Magic (fire spells work best). Rex is weak to Magic and uses melee — Protect from Melee when nearby",
+        "description": "Kill Dagannoth Rex. Use Magic (fire spells work best). Rex is weak to Magic and uses melee - Protect from Melee when nearby",
         "worldX": 2521,
         "worldY": 3740,
         "worldPlane": 0,
@@ -3504,7 +3504,7 @@
         "completionDistance": 20
       },
       {
-        "description": "Kill Dagannoth Prime. Use Ranged. Prime attacks with Magic — use Protect from Magic when targeting Prime",
+        "description": "Kill Dagannoth Prime. Use Ranged. Prime attacks with Magic - use Protect from Magic when targeting Prime",
         "worldX": 2521,
         "worldY": 3740,
         "worldPlane": 0,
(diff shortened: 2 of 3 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#507
**Files changed**: 3 (+84/-41)
**Reviewer focus**: lintDropRates correctness + encoding handling, em-dash replacement readability, charset test inversion preserves issue #433 regression coverage.

### Findings

#### CRITICAL (0)
- None.

#### HIGH (0)
- None.

#### MEDIUM (1)
- `lintDropRates` uses `eachLine('UTF-8')` correctly (verified: injecting a real UTF-8 em-dash byte sequence reports `U+2014` as expected, not the cp1252 mojibake decoding `U+00E2`). The encoding parameter is the load-bearing detail here on Windows — worth a brief inline comment to preserve intent if the task ever gets refactored.

#### LOW (2)
- Lint error message renders the violating glyph after the U+ codepoint (e.g. `U+2014 ('—')`), which on cp1252 consoles will print `?`. Not a bug — the codepoint is the source of truth — but `(after U+XXXX = "…")` could be omitted entirely on grounds that the codepoint alone is unambiguous.
- `violations.take(20)` caps the printed list to 20 entries, but the inner `for` loop already `break`s after the first violation on a line. Combined, this means worst case 20 lines × 1 violation = 20 entries, which is fine — just noting the two limits compose somewhat redundantly.

### Verification ran
- [x] `./gradlew build` — BUILD SUCCESSFUL (lintDropRates, jacocoTestCoverageVerification, and 31 DropRateDatabaseTest tests all green; lintDropRates executes inside `check`)
- [x] `./gradlew lintDropRates` on PR HEAD — passes
- [x] Negative test: injected `\xe2\x80\x94` UTF-8 bytes into line 1142 — lint fails with `line 1142 col 125: codepoint U+2014`, exactly as documented
- [x] `./gradlew check --dry-run` lists `:lintDropRates` before `:check` — wiring confirmed
- [x] `perl -ne 'print if /[^\x00-\x7f]/' drop_rates.json` — 0 hits
- [x] `perl -ne 'print if /\xc3\xa2/' drop_rates.json` (cp1252 mojibake `â` sequences) — 0 hits
- [x] Spot-checked all 18 replaced lines — every one reads naturally with `-`, no `"X--Y"` artifacts, no doubled spaces, no broken word breaks
- [x] Old `foundAtLeastOneEmDash` assertion is fully removed from `DropRateDatabaseTest` (no leftover opposite invariant)
- [x] Inverted test still guards #433: it asserts `MOJIBAKE_PREFIX` (U+00E2 U+20AC, the cp1252 misread of UTF-8 em-dash bytes) is absent in every loaded description and `perItemStepDescription` override — original regression coverage preserved and tightened
- [x] Commit subject, branch name `ci/501-drop-rates-ascii-lint`, PR title all conform to the conventional commit format
- [x] No Claude attribution in commit or PR body
- [x] Author is `cha-ndler <48898494+cha-ndler@users.noreply.github.com>`

### Recommendation
- **SHIP** — no CRITICAL/HIGH issues. The lint is encoding-correct, wired into `check`, fails informatively on regression, and passes on the cleaned file. The test inversion preserves the #433 mojibake guard and additionally enforces the new ASCII-only policy on the loaded model (catching drift via `perItemStepDescription` overrides that the file-level lint wouldn't see if introduced through code-generation).

### Notes
- Nice belt-and-suspenders: file-level lint blocks raw-byte regressions on the JSON file itself, and the unit test extends the invariant to the loaded `CollectionLogSource` model. Future drift via programmatic loaders or generator code is still caught.
- The "three things in one PR" scoping is sensible here — the lint would have failed on master if landed first, so atomic delivery (lint + content fix + test inversion) is the only correct ordering.
- Suggest a small follow-up issue (optional): extend `lintDropRates` to also flag a curated set of fancy-punctuation codepoints (smart quotes U+2018/2019/201C/201D, ellipsis U+2026, non-breaking space U+00A0) explicitly in the violation message, so an author who pastes one of those gets a precise "use ASCII X instead" hint rather than the generic policy line. Current implementation already catches them via the > 0x7F gate; this would just upgrade the diagnostic.

Issue #503, opened 2026-05-17
refactor(tracking): split god-classes >850 LOC into focused sub-coordinators


cha-ndler:
## Problem

Four files in the codebase exceed 850 LOC each:

| File | LOC |
| --- | ---: |
| `src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java` | 1,184 |
| `src/main/java/com/collectionloghelper/guidance/GuidanceOverlayCoordinator.java` | 1,113 |
| `src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java` | 900 |
| `src/main/java/com/collectionloghelper/guidance/GuidanceSequencer.java` | 868 |

Total: 4,065 lines across four files that each individually exceed the working memory of any human reader.

This is a **tracking issue** for splitting these into focused sub-coordinators. Each child PR lands one extraction at a time so review stays manageable and the build/coverage gate stays green.

### Plugin.java
- [ ] `E-a` Extract `ChatEventHandler` (currently lines ~601-640) — chat parsing, regex matching, item notifications, `/clh` command router.
- [ ] `E-b` Extract `LootSyncManager` (currently lines ~641-717) — `onNpcLootReceived`, `onItemContainerChanged`, inventory/bank sync.

### GuidanceOverlayCoordinator.java
- [ ] `E-c` Extract `OverlayStepApplier` from `applyStepToOverlays()` (currently lines ~565-744, ~180 LOC). **This is the largest single seam in the codebase and should land first.**
- [ ] `E-d` Extract `DynamicTargetManager` from `tickDynamicTargetEvaluator()` (currently lines ~450-492).
- [ ] `E-e` Extract `WorldMapController` from world-map-arrow/destination/route logic (currently lines ~988-1046).

### GuidanceSequencer.java
- [ ] `E-f` Extract `CompletionChecker` from completion-condition evaluation (currently lines ~369-478).
- [ ] `E-g` Extract `StepAdvancer` from `skipStep`/`advanceStep`/`skipSatisfiedSteps` (currently lines ~658-783).

### CollectionLogHelperPanel.java
- [ ] `E-h` Extract `PanelModeDispatcher` from `rebuild()` (currently lines ~589-670).
- [ ] `E-i` Extract `StepProgressRenderer` from the three `updateStepProgress()` overloads (currently lines ~774-839).

## Constraints (apply to every child PR)

- One extraction per PR
- `./gradlew build` and `./gradlew test` must pass
[...]

PR #510, opened 2026-05-18
refactor: extract OverlayStepApplier from GuidanceOverlayCoordinator (#503)

cha-ndler:
## Summary

Pure extraction of `GuidanceOverlayCoordinator.applyStepToOverlays()` (~173 LOC) into a new dedicated class `OverlayStepApplier` in `com.collectionloghelper.guidance`. The applier is stateless: coordinator-owned mutable state (`activeTargetItemId`, `lastMessagedStepIndex`, `activeMapPoint`, `collectionLogIcon`) is passed in via `Input` and the updated values are returned via `Result`. The deferred `pendingShortestPathTarget` assignment preserves its original client-thread timing via a `Consumer` callback. Coordinator-local helpers `shouldSetHintArrowTo` / `resolveStepIconType` stay on the coordinator (still used by `applySourceToOverlays`, `tickDynamicTargetEvaluator`, `updateWorldMapArrow`) and are passed in as method references.

`GuidanceOverlayCoordinator.java` shrinks from **1113 LOC** to **949 LOC**. Four reflective-constructor tests are updated to add `OverlayStepApplier` to the constructor signature; each builds a real applier from the same overlay mocks the test already provides so the existing verifications continue to observe the same `setX()` calls.

Partially addresses #503.

## Test plan

- [x] `./gradlew test` passes (1172 tests, 0 failures, 0 ignored); JaCoCo 45% gate holds
- [x] `GuidanceOverlayCoordinator.java` LOC drops from **1113** to **949** (under the 950 target)
- [x] `OverlayStepApplier` wires all overlay/dep references the original method used (9 overlays + `Client`, `ClientThread`, `EventBus`, `Config`, `GuidanceSequencer`, `PlayerTravelCapabilities`, `RequiredItemResolver`, `WorldMapPointManager`)
- [x] Behavioral diff: zero -- pure extraction; `setX()` call ordering, deferred `clientThread.invokeLater` blocks, and `pendingShortestPathTarget` write timing preserved verbatim

### DIFF
--- a/src/main/java/com/collectionloghelper/guidance/GuidanceOverlayCoordinator.java
+++ b/src/main/java/com/collectionloghelper/guidance/GuidanceOverlayCoordinator.java
@@ -37,7 +37,6 @@ import com.collectionloghelper.data.PlayerInventoryState;
 import com.collectionloghelper.data.PlayerTravelCapabilities;
 import com.collectionloghelper.data.RequirementRow;
 import com.collectionloghelper.data.RequirementsChecker;
-import com.collectionloghelper.guidance.DynamicTargetEvaluator;
 import com.collectionloghelper.guidance.dynamic.DynamicTargetEvaluatorRegistry;
 import com.collectionloghelper.overlay.CollectionLogWorldMapPoint;
 import com.collectionloghelper.overlay.DialogHighlightOverlay;
@@ -114,6 +113,7 @@ public class GuidanceOverlayCoordinator
 	private final WorldMapDestinationOverlay worldMapDestinationOverlay;
 	private final GroundItemHighlightOverlay groundItemHighlightOverlay;
 	private final WidgetHighlightOverlay widgetHighlightOverlay;
+	private final OverlayStepApplier overlayStepApplier;
 
 	// -- Guidance UI state (previously in plugin) --
 
@@ -187,7 +187,8 @@ public class GuidanceOverlayCoordinator
 		WorldMapRouteOverlay worldMapRouteOverlay,
 		WorldMapDestinationOverlay worldMapDestinationOverlay,
 		GroundItemHighlightOverlay groundItemHighlightOverlay,
-		WidgetHighlightOverlay widgetHighlightOverlay)
+		WidgetHighlightOverlay widgetHighlightOverlay,
+		OverlayStepApplier overlayStepApplier)
 	{
 		this.client = client;
 		this.clientThread = clientThread;
@@ -211,6 +212,7 @@ public class GuidanceOverlayCoordinator
 		this.worldMapDestinationOverlay = worldMapDestinationOverlay;
 		this.groundItemHighlightOverlay = groundItemHighlightOverlay;
 		this.widgetHighlightOverlay = widgetHighlightOverlay;
+		this.overlayStepApplier = overlayStepApplier;
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#510
**Files changed**: 6  (+477/-182)
**Reviewer focus**: behavioral equivalence of the extraction, threading semantics for `clientThread.invokeLater` blocks and the deferred `pendingShortestPathTarget` write, and whether the `Input`/`Result` indirection introduced any subtle state-staleness.

### Findings

#### CRITICAL (0)
_None._

#### HIGH (0)
_None._

#### MEDIUM (1)
- **`OverlayStepApplier` DI surface is large (17 constructor deps)** — the extraction is faithful (it consumed exactly what the original method consumed), but the new class inherits the same "wires every overlay" smell the coordinator had. Not a blocker for this PR (#503 is an iterative split), but worth a follow-up grouping (e.g., a `GuidanceOverlayBundle` holder, or splitting into "world-location apply" vs "filter/highlight apply" subroutines) before further extraction PRs land on the same surface.

#### LOW (2)
- `apply()` is package-private with no `@VisibleForTesting` marker; current tests reach it transitively via the coordinator. Fine for now, but a direct unit test on `OverlayStepApplier.apply()` would let the no-location branch (currently uncovered, see coverage note) be exercised in isolation and would lock the extraction contract.
- `Input` and `Result` are nested package-private classes with `final` fields and a constructor; consider Lombok `@Value` for symmetry with other carrier types in this package.

### Verification ran
- [x] `./gradlew build` — BUILD SUCCESSFUL
- [x] `./gradlew test` — 1172/1172 passed, 0 failures, 0 skipped, JaCoCo 45% gate held
- [x] Semantic diff of original `applyStepToOverlays` (master `c6cd88a3` lines 565-737) vs new `OverlayStepApplier.apply` — identical `setX()` calls in identical order; both `clientThread.invokeLater(...)` blocks (lines 573, 645, 684, 725 in the original) preserved verbatim, including the inner branching on `config.useShortestPath()` and `shouldSetHintArrowTo`
- [x] Deferred-write timing for `pendingShortestPathTarget` confirmed equivalent — original assigned the field inside the third `invokeLater` block; new design invokes `pendingShortestPathTargetSetter.accept(hintTarget)` at the same point inside the same `invokeLater` block, with the setter `wp -> pendingShortestPathTarget = wp`. Same thread, same moment
- [x] Coordinator-local helpers `shouldSetHintArrowTo` and `resolveStepIconType` retained on the coordinator (still called from `applySourceToOverlays`, `tickDynamicTargetEvaluator`, `updateWorldMapArrow` at lines 476/485/675/901) and passed to the applier as method references — no duplication
- [x] `GuidanceOverlayCoordinator.applyStepToOverlays` is now a 6-line delegation (build `Input`, call `apply`, write `Result.lastMessagedStepIndex` and `Result.activeMapPoint` back, then `applyDynamicItemObjectOverlays()`)
- [x] LOC: coordinator 1113 → 949 (under the 950 target); applier 375
- [x] JaCoCo coverage on `OverlayStepApplier`: 50% instructions / 19% branches / 116 of 166 lines covered. The uncovered surface is the no-location `else` branch and the inner bodies of the `clientThread.invokeLater` lambdas (which the existing coordinator tests also never exercised — the original had the same coverage profile at 47% instructions). `Input` and `Result` are at 100%. No regression
- [x] Test updates inspected — each of the four reflective-constructor tests adds an `OverlayStepApplier` parameter and builds a **real** applier from the **same** mocks the test already wires; existing `verify(mockOverlay).setX(...)` assertions continue to fire because they observe calls the applier now makes on those same mocks. No mock-stub gymnastics, no assertions weakened
- [x] No orphan helpers, no dead imports, no scope creep

### Recommendation
**SHIP** — pure extraction is genuinely pure. Behavior, threading, and call ordering all preserved. Tests carry over cleanly with their assertions intact. Coordinator clears the <950 target. The MEDIUM is a forward-looking note, not a blocker.

### Notes
- The `Input` / `Result` plus `Consumer` design is a sound pattern for "keep mutation ownership on the parent, move computation to the child." Worth reusing for the next splits under #503 — flag it in the issue as the canonical shape so subsequent extractor PRs don't reinvent it.
- The `Input.activeTargetItemId` / `Input.collectionLogIcon` snapshot is safe **because** `apply()` runs synchronously to completion; if any future caller of `apply()` ever defers it (e.g., wraps the whole call in `invokeLater`), the snapshot semantics will diverge from "read field at use time." Worth a one-line javadoc on `Input` saying "must be constructed and passed synchronously into `apply()`; do not reuse across deferred boundaries."
- Suggested checklist addition (not in scope for this PR, but flagging for the reviewer template): for god-class-split PRs, verify that helper methods retained on the parent remain called from at least one non-extracted path — caught here by inspection but worth automating.

PR #513, opened 2026-05-18
refactor: extract PanelRebuildOrchestrator from Panel (#503)

cha-ndler:
## Summary

Extracts the scroll-position + expanded-category snapshot/restore logic out of `CollectionLogHelperPanel.rebuild()` into a new `PanelRebuildOrchestrator` helper. The panel's rebuild body now reads as a short orchestration: capture state, tear down `listContainer`, let `modeDispatcher.buildView()` repopulate it, restore expanded categories, defer scroll restore. Behaviour is unchanged; thread discipline (the outer `SwingUtilities.invokeLater` and the `rebuilding` / `rebuildPending` state machine) remains on the panel where it belongs.

Partially addresses #503.

### Changes

- **New** `src/main/java/com/collectionloghelper/ui/PanelRebuildOrchestrator.java` (175 LOC)
  - `capture()` snapshots scroll position (via the enclosing `JScrollPane` ancestor) and the set of currently-expanded `CategorySummaryPanel` names in `listContainer`.
  - `restoreExpanded(snapshot)` re-applies the expanded set to whichever category panels are present after rebuild, looked up by name.
  - `deferScrollRestore(snapshot)` schedules the scroll restore via `SwingUtilities.invokeLater` so the new component tree has a chance to lay out before the offset is re-applied. No-ops cleanly when there is no scroll pane or the saved offset is zero.
  - Immutable nested `RebuildSnapshot` value holder (Java 11 target, so a plain final-field class, not a record).
- **Modified** `src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java`
  - `rebuild()` body shrinks from 80 LOC to 41 LOC.
  - Panel total LOC: 900 → 861 (−39).
  - New `private final PanelRebuildOrchestrator rebuildOrchestrator` field, constructed in the panel constructor after `listContainer` is created.
  - Removed unused `java.util.HashSet` / `java.util.Set` imports.
- **New** `src/test/java/com/collectionloghelper/ui/PanelRebuildOrchestratorTest.java` (11 tests)

### LOC

| File | Before | After | Delta |
|---|---|---|---|
| `CollectionLogHelperPanel.java` | 900 | 861 | −39 |
| `PanelRebuildOrchestrator.java` | — | 175 | new |

## Test plan

- [x] `./gradlew build` passes (includes `jacocoTestCoverageVerification`)
- [x] `./gradlew test` — 1183 tests, 0 failures, 0 errors, 0 skipped (+11 new tests on the orchestrator)
- [x] New unit tests cover:
  - capture returns zero scroll + empty expanded set when nothing is expanded and the scrollbar is at 0
  - capture records non-zero scroll position from the enclosing `JScrollPane`
  - capture records only currently-expanded category names (not collapsed or non-category children)
  - `restoreExpanded` re-expands matching categories after a rebuild that recreates the same names
  - `restoreExpanded` is a no-op on an empty snapshot
  - `restoreExpanded` does not re-expand a category whose name is absent from the snapshot
  - `deferScrollRestore` schedules a value-set on the vertical scrollbar via `invokeLater`
  - `deferScrollRestore` no-ops when the saved offset is 0
  - `deferScrollRestore` no-ops when the panel is not inside a scroll pane
  - the snapshot's `expandedCategories` set is unmodifiable
- [x] Existing panel tests under `src/test/java/com/collectionloghelper/ui/` continue to pass; panel public API unchanged
- [x] No new `SwingUtilities.invokeLater` wrapping introduced beyond the original (`deferScrollRestore` matches the original inline `invokeLater` for the deferred scroll restore exactly)

### DIFF
--- a/src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java
+++ b/src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java
@@ -60,9 +60,7 @@ import com.collectionloghelper.ui.widget.SlayerStrategyView;
 import com.collectionloghelper.ui.widget.StepProgressView;
 import com.collectionloghelper.ui.widget.SyncStatusView;
 import java.util.EnumMap;
-import java.util.HashSet;
 import java.util.Map;
-import java.util.Set;
 import java.awt.BorderLayout;
 import java.awt.CardLayout;
 import java.awt.Color;
@@ -173,6 +171,7 @@ public class CollectionLogHelperPanel extends PluginPanel implements PanelShellC
 
 	private final Map<Mode, PanelModeController> modeControllers = new EnumMap<>(Mode.class);
 	private final PanelModeDispatcher<Mode> modeDispatcher = new PanelModeDispatcher<>(modeControllers);
+	private final PanelRebuildOrchestrator rebuildOrchestrator;
 
 	private Mode currentMode = Mode.EFFICIENT;
 	private boolean rebuilding = false;
@@ -420,6 +419,7 @@ public class CollectionLogHelperPanel extends PluginPanel implements PanelShellC
 		listContainer.setLayout(new BoxLayout(listContainer, BoxLayout.Y_AXIS));
 		listContainer.setBackground(ColorScheme.DARK_GRAY_COLOR);
 		listView.add(listContainer, BorderLayout.NORTH);
+		rebuildOrchestrator = new PanelRebuildOrchestrator(this, listContainer);
 
 		detailView = new JPanel(new BorderLayout());
 		detailView.setBackground(ColorScheme.DARK_GRAY_COLOR);
@@ -601,28 +601,7 @@ public class CollectionLogHelperPanel extends PluginPanel implements PanelShellC
 			rebuilding = true;
 			try
 			{
-				// Save scroll position before rebuild
-				int savedScrollPosition = 0;
-				JScrollPane scrollPane = (JScrollPane) SwingUtilities.getAncestorOfClass(
-					JScrollPane.class, this);
-				if (scrollPane != null)
-				{
-					savedScrollPosition = scrollPane.getVerticalScrollBar().getValue();
-				}
-
-				// Save expanded category names before clearing
-				Set<String> expandedCategories = new HashSet<>();
-				for (Component comp : listContainer.getComponents())
-				{
-					if (comp instanceof CategorySummaryPanel)
-					{
-						CategorySummaryPanel csp = (CategorySummaryPanel) comp;
-						if (csp.isExpanded())
-						{
-							expandedCategories.add(csp.getCategoryName());
-						}
-					}
-				}
+				PanelRebuildOrchestrator.RebuildSnapshot snapshot = rebuildOrchestrator.capture();
 
 				SwingUtil.fastRemoveAll(listContainer);
 				updateCompletionHeader();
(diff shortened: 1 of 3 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#513
**Files changed**: 3  (+467/-44)
**Reviewer focus**: EDT discipline preservation, pure-extraction equivalence against the original `rebuild()` body, and test-quality spot-check on the 11 new orchestrator tests.

### Findings

#### CRITICAL (0)
- None.

#### HIGH (0)
- None.

#### MEDIUM (0)
- None.

#### LOW (1)
- `PanelRebuildOrchestrator.java` is 175 LOC for ~80 LOC of actual logic (the rest is the BSD license header + Javadoc). Acceptable for project conventions but on the verbose side; if a future change adds more orchestration concerns, consider whether the file is still cohesive enough.

### Verification ran
- [x] `./gradlew test` (rerun-tasks): BUILD SUCCESSFUL, 1183 tests, 0 failures, 0 errors, 0 skipped.
- [x] Orchestrator test class: 11 tests, all passing.
- [x] LOC counts confirmed: Panel 900 → 861 (−39); Orchestrator 175 (new); Test file 287 (new).
- [x] Outer `SwingUtilities.invokeLater` wrap around `rebuild()` body preserved on the panel (line 593). No second `invokeLater` introduced — `deferScrollRestore` reuses the original inline `invokeLater` pattern, just relocated.
- [x] `rebuilding` / `rebuildPending` state machine remains on `CollectionLogHelperPanel` (lines 595-599, 624-629). Not leaked to orchestrator.
- [x] `inDetailView` gate around `showListView()` + `deferScrollRestore()` preserved on the panel (lines 616-620).
- [x] `capture()` body is a byte-for-byte equivalent of the original snapshot block — same `SwingUtilities.getAncestorOfClass(JScrollPane.class, ...)` lookup, same `CategorySummaryPanel` instanceof filter, same `getCategoryName()` keying.
- [x] `restoreExpanded()` body matches the original expanded-state restore loop.
- [x] `deferScrollRestore()` reuses the snapshot's cached `scrollPane` reference (matches the original lambda capture of the local `scrollPane` variable — no re-lookup at restore time, so semantics are identical even if the ancestor changes mid-rebuild).
- [x] No additional state captured (no focus / selection) and none dropped. Pure extraction.
- [x] No new fields introduced on the panel beyond `rebuildOrchestrator`; removed `HashSet` / `Set` imports are no longer referenced on the panel side.

### Test quality spot-check (4 of 11)
- `captureRecordsScrollPositionFromEnclosingScrollPane` — sets vbar to 250, asserts snapshot equals 250. Real behavioral assertion.
- `restoreExpandedReExpandsMatchingCategoriesAfterRebuild` — strongest test: expands `bosses`, captures, simulates rebuild by removing and re-adding **new** `CategorySummaryPanel` instances with the same names, then asserts the new instance is re-expanded and an unrelated category remains collapsed. Verifies the name-based reattachment correctly bridges old → new component identity.
- `deferScrollRestoreSchedulesValueOnVerticalScrollBar` — captures at 400, resets vbar to 0, defers restore, flushes EDT via `invokeAndWait`, asserts vbar == 400. Verifies the deferred restore actually fires.
- `deferScrollRestoreIsNoOpWhenScrollPositionIsZero` — captures at 0, sets vbar to 123, defers, asserts vbar still 123 after flush. Verifies the zero-guard branch.

No tautological assertions found; no \"verify mock called N times\" without a behavioral check.

The only marginal test is `deferScrollRestoreIsNoOpWhenPanelNotInScrollPane`, which can only assert \"no exception thrown\" and `scrollPosition == 0` (asserting an `invokeLater` was *not* scheduled is awkward from outside Swing). Acceptable.

### Three-method API vs single-callback shape
The agent's deviation from the suggested `rebuild(Runnable)` shape is well-justified by the actual structure of the original `rebuild()` body: there is work *between* `restoreExpanded` (called before `revalidate`/`repaint`) and `deferScrollRestore` (called only inside the `!inDetailView` branch, after `showListView()`). A single-callback shape would either have required reshuffling the `inDetailView` gate or smuggling Swing-state through closures — both worse than the explicit three-step API. The snapshot is passed by argument across all three calls and lives only in `rebuild()`'s local scope, so there is no orchestrator-internal cross-call state to fall out of sync.

### Recommendation
**SHIP** — no CRITICAL/HIGH/MEDIUM issues, all checklist items pass, behavior provably preserved, tests have real behavioral assertions, partially addresses #503 as advertised.

### Notes
- The `RebuildSnapshot` value holder exposes `getScrollPosition()` and `getExpandedCategories()` publicly but **not** `scrollPane`. That is the right choice — `scrollPane` is an implementation detail used only by `deferScrollRestore`, and hiding it preserves encapsulation while still allowing the public getters for testing.
- The agent's docblock on the orchestrator includes the expected call sequence, which is useful documentation for the contract between the panel and the helper.
- Suggested follow-up (not blocking): if/when more rebuild concerns are extracted from the panel (the rest of #503's god-class split), consider whether `PanelRebuildOrchestrator` should grow to own the `rebuilding`/`rebuildPending` re-entrancy gate too. Today it must stay on the panel because the panel owns the outer `invokeLater`; if the orchestrator ever takes that over, the gate could move with it.

PR #520, opened 2026-05-18
refactor: extract WorldMapController from GuidanceOverlayCoordinator (#503)

cha-ndler:
## Summary

Continues the god-class breakup tracked in #503. Extracts the world-map / hint-arrow plumbing cluster out of `GuidanceOverlayCoordinator.java` into a new `@Singleton` helper `WorldMapController.java`, mirroring the DI idiom established by `OverlayStepApplier` in #510.

Methods moved (all formerly private to the coordinator):

- `updateArrow(CollectionLogWorldMapPoint)` (renamed from `updateWorldMapArrow`) — rotates the active map point's edge-snap arrow each tick.
- `shouldSetHintArrowTo(WorldPoint)` — pure predicate gating `client.setHintArrow` on `showHintArrow`, top-level `WorldView`, matching plane.
- `refreshHintArrow(boolean isActive, GuidanceStep step)` — clears + re-applies the hint arrow on the client thread. Coordinator state now passes through args instead of via a back-reference.
- `resolveStepIconType(GuidanceStep, Integer activeTargetItemId)` — maps a step to a `WorldMapDestinationOverlay.StepIconType`.

`GuidanceOverlayCoordinator` keeps a one-line public `refreshHintArrow()` delegating to the controller, preserving the external API contract used by `GuidanceMovementTracker` and `GuidanceEventRouter` (and the `verify(coordinator).refreshHintArrow()` assertions in their tests).

LOC: `GuidanceOverlayCoordinator` 949 → 843 (−106). New `WorldMapController` 210.

## Test plan

- [x] `./gradlew build` green (compile + test + JaCoCo verification + lintDropRates).
- [x] Existing `GuidanceMovementTrackerTest` and `GuidanceEventRouterTest` pass unmodified — public `refreshHintArrow()` signature unchanged.
- [x] Reflective-constructor tests updated minimally to add `WorldMapController.class` to the coordinator constructor signature: `GuidanceOverlayCoordinatorMapPointTest`, `GuidanceOverlayCoordinatorDescriptionTest`, `GuidanceOverlayCoordinatorNpcIdTest`, `DynamicTargetEvaluatorDispatchTest`.
- [x] JaCoCo 45% coverage gate holds.
- [ ] Manual smoke test in RuneLite: activate guidance on a step with `worldX > 0`, confirm hint arrow appears and follows player on teleport.
- [ ] Manual smoke test: toggle "Show Hint Arrow" mid-guidance, confirm arrow clears/reappears immediately.
[...]

PR #516, opened 2026-05-18
refactor: extract StepAdvancer from GuidanceSequencer (#503)

cha-ndler:
## Summary

Wave 4 god-class split for `GuidanceSequencer.java` — extracts the step-advancement cluster (`skipStep`, `advanceStep`, `skipSatisfiedSteps`, plus the `finishStepTransition` helper introduced in #512) into a new `@Singleton StepAdvancer`. Follows the Input/Result carrier pattern established by `OverlayStepApplier` (#510) and `CompletionChecker` (#512).

- `StepAdvancer.runSkipChain(SkipChainInput)` — the "skip already-satisfied steps" scan
- `StepAdvancer.skip(TransitionInput)` — full Skip-button plan (`currentIndex++`, reset loop counter, run skip-chain)
- `StepAdvancer.advance(TransitionInput)` — full Next-step plan (loop-back vs forward, then skip-chain)

The advancer holds no mutable state. It returns a `TransitionOutcome` (newIndex, newLoopIterationsCompleted, sequenceComplete, loopBackResume); the sequencer applies the mutation in a small `applyTransition()` helper and fires the existing `onSequenceComplete` / `notifyStepChanged` callbacks. `finishStepTransition()` folds into `applyTransition()`, and the unused `isStepAlreadySatisfied()` private helper is removed — `CompletionChecker` is invoked directly from inside `StepAdvancer.runSkipChain`.

`GuidanceSequencer.java`: **750 -> 692 LOC** (-58)
`StepAdvancer.java`: **313 LOC** (new)

Partially addresses #503.

## Test plan

- [x] `./gradlew build` succeeds (compile + JaCoCo 45% coverage gate)
- [x] `./gradlew test` — 1183 tests, 0 failures, 0 errors, 0 skipped
- [x] `GuidanceSequencerTest` (67 cases), `GuidanceSequencerNestedConditionalTest` (12 cases), `GuidanceSequencerWaypointTest` (6 cases) all pass without modification — the carrier pattern preserves the public API and observable behaviour (loop-back resume, skip-chain past final step, `onSequenceComplete` firing at most once)
- [x] Existing `@Inject` constructor signature for `GuidanceSequencer` is unchanged — reflection-based test construction continues to work; `StepAdvancer` is built inside the constructor alongside the existing inline `CompletionChecker`
- [x] No public API changes on `GuidanceSequencer`; all 22 production + test callers compile unchanged

PR #512, opened 2026-05-18
refactor: extract CompletionChecker from GuidanceSequencer (#503)

cha-ndler:
Partially addresses #503.

## Summary

Splits the completion-condition evaluation cluster out of the 868-LOC
`GuidanceSequencer` god-class into a new `CompletionChecker` collaborator.

- `GuidanceSequencer.java`: 868 -> 750 LOC
- New file: `CompletionChecker.java` (445 LOC, includes license header + Input/Result static nested classes)

### Extracted to CompletionChecker

Pure predicates (no state, no side effects):
- `isItemObtainedSatisfying`
- `isInventoryHasItemSatisfying` / `isInventoryNotHasItemSatisfying`
- `evaluateSkipIfHasAny`
- `isNpcDeathSatisfying` / `isNpcInteractedSatisfying`
- `isVarbitAtLeastSatisfying`
- `isArriveAtZoneSatisfying` / `isPlayerOnPlaneSatisfying`
- `isStepAlreadySatisfied` (skip-chain helper)

Input/Result-style methods (sequencer retains cache ownership):
- `evaluateArriveAtTile` returns updated `TilePointCache`
- `evaluateChatMessage` returns updated `ChatPatternCache`
- `evaluateWaypointProgress` returns updated `crossedWaypointIndex`

### Sequencer responsibilities preserved

`GuidanceSequencer` continues to own all mutable state — `currentIndex`,
`loopIterationsCompleted`, `crossedWaypointIndex`, tile/chat caches,
`resolvedAlternatives` — and remains responsible for step advancement,
loop iteration, and `onSequenceComplete` callbacks. The new collaborator
performs only pure evaluation; callers inspect the returned boolean (and
updated cache) and apply state transitions themselves.

### DI / API compatibility

Public methods of `GuidanceSequencer` are unchanged. The 4-arg `@Inject`
constructor signature is preserved (CompletionChecker is built inline
from the same dependencies inside the constructor body) so the
reflection-based test setups in `GuidanceSequencerTest`,
`GuidanceSequencerWaypointTest`,
`GuidanceSequencerNestedConditionalTest`, `BossGuidanceDispatchTest`,
etc. continue to work without modification.

### Additional cleanup (within scope)

- New `finishStepTransition()` helper deduplicates the post-skip-chain
  tail logic shared between `skipStep()` and `advanceStep()`.
- `getRawCurrentStep()` now delegates to `getCurrentStep()` (they had
  identical bodies after the bank-routing substitution was removed
[...]

PR #523, opened 2026-05-18
refactor: extract DynamicTargetManager from GuidanceOverlayCoordinator (#503)

cha-ndler:
## Summary

Partially addresses #503.

Extracts the per-tick dynamic-target evaluator dispatch out of `GuidanceOverlayCoordinator` into a new `DynamicTargetManager`, continuing the coordinator decomposition started by #510 (`OverlayStepApplier`) and #520 (`WorldMapController`).

- `DynamicTargetManager` is `@Singleton`, constructor `@Inject`, stateless — coordinator state it needs is passed in via a static `Input` carrier and any state mutation (the new `activeMapPoint`) is returned in a `Result` so the coordinator remains the sole owner of guidance state.
- The `DynamicTargetEvaluatorRegistry` dependency moves with the dispatch logic; `GuidanceOverlayCoordinator` no longer imports or references it.
- `tick()` in the coordinator now builds `DynamicTargetManager.Input`, calls `manager.tick(input)`, and assigns the result's `activeMapPoint`.

LOC: `GuidanceOverlayCoordinator` 843 -> 792 (-51). `DynamicTargetManager` is 234 LOC.

## Test plan

- [x] `./gradlew build` (compileJava + compileTestJava + test + jacocoTestCoverageVerification + lintDropRates) passes locally.
- [x] All 1200 tests green; previously-affected reflective-constructor tests now wire the new ctor signature.
- [x] JaCoCo 45% coverage gate holds.
- [x] No production behaviour change: the same overlay setters and `WorldMapPointManager.add/remove` calls fire in the same order for the same inputs.
- [x] `DynamicTargetEvaluatorDispatchTest` (the behavioural test for this seam) passes against the new wiring — exercises `tick_*` paths for active/inactive, null/unknown key, null-evaluator-result.

### DIFF
--- a/src/main/java/com/collectionloghelper/guidance/DynamicTargetManager.java
+++ b/src/main/java/com/collectionloghelper/guidance/DynamicTargetManager.java
@@ -1,0 +1,234 @@
+/*
+ * Copyright (c) 2025, cha-ndler
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.collectionloghelper.guidance;
+
+import com.collectionloghelper.data.GuidanceStep;
+import com.collectionloghelper.guidance.dynamic.DynamicTargetEvaluatorRegistry;
+import com.collectionloghelper.overlay.CollectionLogWorldMapPoint;
+import com.collectionloghelper.overlay.GuidanceMinimapOverlay;
+import com.collectionloghelper.overlay.GuidanceOverlay;
+import com.collectionloghelper.overlay.WorldMapDestinationOverlay;
+import com.collectionloghelper.overlay.WorldMapRouteOverlay;
+import java.awt.image.BufferedImage;
+import javax.annotation.Nullable;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import lombok.extern.slf4j.Slf4j;
+import net.runelite.api.Client;
+import net.runelite.api.coords.WorldPoint;
+import net.runelite.client.ui.overlay.worldmap.WorldMapPointManager;
+
+/**
+ * Owns the per-tick dynamic-target evaluator dispatch for guidance.
+ *
+ * <p>Pure extraction from {@link GuidanceOverlayCoordinator}: this manager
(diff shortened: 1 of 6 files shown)


cha-ndler (commented):
@-

cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#523
**Files changed**: 6 (+318/-75)
**Reviewer focus**: Pure-extraction behavior parity, world-map double-application risk, per-tick allocation, registry-move cleanliness, reflective-test wiring.

### Findings

#### CRITICAL (0)
- none

#### HIGH (0)
- none

#### MEDIUM (1)
- **Per-tick allocations on guard-fail paths**: `DynamicTargetManager.tick()` is called unconditionally from `GuidanceOverlayCoordinator.tick()`, which means `new DynamicTargetManager.Input(...)` is allocated every game tick even when guidance is inactive or the step has no evaluator key. Master's `tickDynamicTargetEvaluator()` did the same unconditional call but allocated nothing on the early-return paths. Added cost per tick: 1 `Input` + 1 `Result.unchanged(...)` object. Not a regression worth blocking on (small short-lived objects), but flagging given the project's overlay-performance discipline. Two cheap mitigations if you want them: (a) lift the `guidanceSequencer.isActive()` guard back into the coordinator so the manager is only invoked when guidance is active, or (b) cache a singleton `Result.unchanged(null)` for the common no-op case. Worth a follow-up issue, not a blocker.

#### LOW (2)
- `DynamicTargetManager.Result` could be a `record` (Java 17) per the project's Java coding-style guidance on modern features. The existing carrier pattern across `OverlayStepApplier`, `CompletionChecker`, `StepAdvancer`, `WorldMapController` appears to use plain final classes for consistency, so keeping the same style is defensible — flagging only as optional polish.
- Each of the 4 reflective tests now contains a near-identical `newDynamicTargetManager(WorldMapController)` helper (~12 lines apiece). Same duplication shape the project already tolerates for `newOverlayStepApplier` / `newWorldMapController`, so this is consistent with surrounding code rather than new debt — could be hoisted to a shared test fixture in a future cleanup.

### Verification ran
- [x] `./gradlew test --rerun-tasks` — **1200 successful, 0 failed, 0 skipped** (4.252s).
- [x] Behavior parity: statement-by-statement diff of master `tickDynamicTargetEvaluator()` vs `DynamicTargetManager.tick(Input)`. Same guard ordering (`guidanceActive` -> step+key null -> registry lookup -> evaluator result null), same overlay setter call order (`guidanceOverlay` -> `guidanceMinimapOverlay` -> `worldMapRouteOverlay` -> `worldMapDestinationOverlay.setTarget`), same `remove(old) -> new CollectionLogWorldMapPoint -> add(new) -> setMapPointActive(true) -> shouldSetHintArrowTo -> setHintArrow` sequence.
- [x] World-map update path: `DynamicTargetManager` owns `worldMapPointManager.add/remove` and `client.setHintArrow`; coordinator's post-call `worldMapController.updateArrow(activeMapPoint)` only computes the rotation degree and calls `activeMapPoint.rotateArrow(...)` — **no double-application** of map-point CRUD or hint-arrow.
- [x] No per-tick allocation in the manager that wasn't already in master's body, **except** the new `Input` and `Result` carrier objects (covered as MEDIUM above). The `new CollectionLogWorldMapPoint(...)` inside the fire path is unchanged from master.
- [x] Registry move clean: grepped `DynamicTargetEvaluatorRegistry` across `src/main/java`. Only references are the registry class itself, the new `DynamicTargetManager` field/import, and unrelated javadoc `@link` references in `DynamicTargetEvaluator.java` / `GuidanceStep.java` / `WintertodtBrazierEvaluator.java`. Coordinator no longer imports or references it.
- [x] Reflective tests consistent: all 4 tests (`DynamicTargetEvaluatorDispatchTest`, `GuidanceOverlayCoordinatorDescriptionTest`, `GuidanceOverlayCoordinatorMapPointTest`, `GuidanceOverlayCoordinatorNpcIdTest`) drop `DynamicTargetEvaluatorRegistry.class` from the ctor signature, append `DynamicTargetManager.class`, and wire a real manager via `newDynamicTargetManager(WorldMapController)` that injects the existing `registry` / `dynamicTargetEvaluatorRegistry` mock — so the test's evaluator-registration stubs still flow into the dispatch path through the manager.
- [x] LOC verified: `GuidanceOverlayCoordinator.java` = **792 lines** (-51 vs master's 843, **short of the <=770 target by 22 LOC**). `DynamicTargetManager.java` = **234 lines**. Inspected the post-PR coordinator for adjacent dynamic-target/world-map code that could have been pulled along — no obvious 25-LOC chunk that clearly belongs with this seam (the remaining tick body is ShortestPath dispatch + `updateArrow` rotation, which is a separate WorldMapController concern, and dragging it in would be scope creep into the #520 seam). Honest trade-off; not flagging.
- [x] No allocations in overlay `render()` paths touched.
- [x] `@Subscribe` discipline untouched.
- [x] `startUp`/`shutDown` overlay symmetry untouched.

### Recommendation
- **SHIP** — no CRITICAL or HIGH issues; one MEDIUM (per-tick guard-fail allocation) that's a small follow-up rather than a blocker, two LOW polish notes. Behavior is byte-for-byte equivalent with master.

### Notes
- The Input/Result carrier pattern is now the established idiom across five seams (#510, #512, #516, #520, #523). Once the coordinator decomposition is complete (post-#503), it would be worth a one-time follow-up to evaluate whether the guard-fail allocations across all five managers sum to something measurable on the per-tick hot path. Cheap to instrument with `-XX:+PrintGCDetails` against a fresh-start session.
- The honest LOC reporting (843 -> 792, short of <=770) is the right call — surfacing the seam's natural size rather than padding the extraction with unrelated code preserves the refactor's reviewability. Adjust the #503 sub-PR ledger expectations rather than the code.
- LOC target shortfall is 22 lines; the next #503 sub-PR (or a coordinator-internal field-grouping pass) can comfortably absorb it.

PR #524, opened 2026-05-18
refactor: extract SequencerEventAdapter from GuidanceSequencer (#503)

cha-ndler:
Partially addresses #503.

## Summary

Lifts the four event-adapter methods out of `GuidanceSequencer` into a dedicated `SequencerEventAdapter`:

- `onNpcDeath(int)` (ACTOR_DEATH)
- `onNpcInteracted(int)` (NPC_TALKED_TO)
- `onChatMessage(String)` (CHAT_MESSAGE_RECEIVED, regex + cache)
- `onVarbitChanged(int, int)` (VARBIT_AT_LEAST)

The adapter owns completion evaluation (via the existing `CompletionChecker` from #512) and constructs the completion log line. The sequencer retains sole ownership of step-index mutation: its public entry points stay on the same signatures, route through a small `EventOutcome` record, and a shared `applyEventOutcome` helper logs and calls `advanceStep()` when the outcome is satisfied.

The chat regex cache is still owned by the sequencer (one cache per active step) and threaded through `onChatMessage` so the adapter stays stateless.

## LOC

| File | Before | After |
| --- | --- | --- |
| `GuidanceSequencer.java` | 692 | 683 |
| `SequencerEventAdapter.java` | -- | 150 (new) |

Smaller sequencer reduction than the original structural map suggested because the event-adapter methods were already thin after #512 (`CompletionChecker`) and #516 (`StepAdvancer`); most of the line count was log strings and `currentIndex+1` references, which moved into the adapter as preformatted log messages.

## Test plan

- [x] `./gradlew build` passes (compileJava, compileTestJava, test, jacocoTestCoverageVerification, lintDropRates, jacocoTestReport, check, build)
- [x] `GuidanceSequencerTest`, `GuidanceSequencerWaypointTest`, `GuidanceSequencerNestedConditionalTest` pass without modification
- [x] `GuidanceEventRouterTest` (verifies `sequencer.onNpcDeath` / `onNpcInteracted`) passes without modification
- [x] JaCoCo 45 percent coverage gate holds
- [x] Public sequencer signatures unchanged; callers in `CollectionLogHelperPlugin`, `GuidanceEventRouter`, and tests need no edits

### DIFF
--- a/src/main/java/com/collectionloghelper/guidance/GuidanceSequencer.java
+++ b/src/main/java/com/collectionloghelper/guidance/GuidanceSequencer.java
@@ -56,6 +56,7 @@ public class GuidanceSequencer
 	private final BossGuidanceRegistry bossRegistry;
 	private final CompletionChecker completionChecker;
 	private final StepAdvancer stepAdvancer;
+	private final SequencerEventAdapter eventAdapter;
 
 	private volatile WorldPoint lastKnownPlayerLocation;
 	private volatile CollectionLogSource activeSource;
@@ -92,6 +93,7 @@ public class GuidanceSequencer
 		this.bossRegistry = bossRegistry;
 		this.completionChecker = new CompletionChecker(inventoryState, collectionState);
 		this.stepAdvancer = new StepAdvancer(this.completionChecker);
+		this.eventAdapter = new SequencerEventAdapter(this.completionChecker);
 	}
 
 	/**
@@ -530,13 +532,7 @@ public class GuidanceSequencer
 		{
 			return;
 		}
-
-		GuidanceStep step = getRawCurrentStep();
-		if (completionChecker.isNpcDeathSatisfying(step, npcId))
-		{
-			log.info("Step {} complete (ACTOR_DEATH: {})", currentIndex + 1, npcId);
-			advanceStep();
-		}
+		applyEventOutcome(eventAdapter.onNpcDeath(getRawCurrentStep(), currentIndex + 1, npcId));
 	}
 
 	/**
@@ -548,17 +544,10 @@ public class GuidanceSequencer
 		{
 			return;
 		}
-
-		GuidanceStep step = getRawCurrentStep();
-		CompletionChecker.ChatMatchResult chatResult =
-			completionChecker.evaluateChatMessage(step, message, chatPatternCache);
-		chatPatternCache = chatResult.cache();
-		if (chatResult.matched())
-		{
-			log.info("Step {} complete (CHAT_MESSAGE_RECEIVED: matched '{}')",
-				currentIndex + 1, step.getCompletionChatPattern());
-			advanceStep();
-		}
+		SequencerEventAdapter.EventOutcome outcome =
+			eventAdapter.onChatMessage(getRawCurrentStep(), currentIndex + 1, message, chatPatternCache);
+		chatPatternCache = outcome.chatCache();
+		applyEventOutcome(outcome);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#524
**Files changed**: 2  (+171/-30)
**Reviewer focus**: pure-extraction correctness of the 4 event-adapter methods, log-message byte-for-byte preservation, chat regex cache threading, and EventOutcome completeness.

### Findings

#### CRITICAL (0)
- none

#### HIGH (0)
- none

#### MEDIUM (0)
- none

#### LOW (1)
- `EventOutcome` is a manual final class with three accessors plus a private constructor; it is exactly the shape that maps onto a Java 16+ `record` (which the project uses elsewhere — e.g. `CompletionChecker.ChatMatchResult`, `CompletionChecker.ChatPatternCache`). Optional follow-up; not blocking.

### Verification ran
- [x] `./gradlew test` — 1205/1205 passed, 0 skipped, 0 failures, 0 errors
- [x] LOC verified: `GuidanceSequencer.java` 692 → 683 (-9); `SequencerEventAdapter.java` = 150 (new); net +141 across both files
- [x] All 4 adapter method bodies match the pre-PR sequencer logic statement-by-statement
  - `onNpcDeath` → `completionChecker.isNpcDeathSatisfying(step, npcId)` (same guard, same arguments)
  - `onNpcInteracted` → `completionChecker.isNpcInteractedSatisfying(step, npcId)`
  - `onVarbitChanged` → `completionChecker.isVarbitAtLeastSatisfying(step, varbitId, value)`
  - `onChatMessage` → `completionChecker.evaluateChatMessage(step, message, cache)`; cache is unconditionally threaded back via `outcome.chatCache()` regardless of `matched`, matching pre-PR `chatPatternCache = chatResult.cache();` semantics (including the early-return path inside `evaluateChatMessage` where the input cache is round-tripped unchanged)
- [x] Log message format preservation: SLF4J `log.info("Step {} complete (ACTOR_DEATH: {})", currentIndex + 1, npcId)` → adapter concatenation `"Step " + n + " complete (ACTOR_DEATH: " + npcId + ")"` then `log.info("{}", msg)` produces the identical final string for all 4 events. `currentIndex + 1` is computed at the call site and passed in as `currentStepNumber`, so step numbering is preserved.
- [x] `EventOutcome` fields cover everything the original methods communicated back to the sequencer's state: `satisfied` (drives `advanceStep`), `completionLogMessage` (the templated `log.info` line), `chatCache` (the only state the chat path mutated). No other state was previously assigned by the four methods.
- [x] `applyEventOutcome` consolidation is correct: on `!satisfied` it returns immediately (matching original early exits); on satisfied it logs then calls `advanceStep()` in the same order as before.
- [x] Sole-owner property holds: `advanceStep()` is still invoked only from the sequencer (the adapter never calls it; it cannot — it has no reference to the sequencer).
- [x] Targeted suites pass unmodified: `GuidanceSequencerTest` (67 tests), `GuidanceSequencerWaypointTest`, `GuidanceSequencerNestedConditionalTest`, `GuidanceEventRouterTest`
- [x] Public sequencer signatures unchanged — `CollectionLogHelperPlugin`, `GuidanceEventRouter`, and tests need no edits
- [x] Member ordering, brace style, indentation match surrounding sequencer code

### Recommendation
- **SHIP** — pure extraction verified, behavior preserved, all tests green. The -9 LOC delta on the sequencer is modest, but the seam is real: the adapter is now the single owner of the four event-condition completion log formats and is the only place that needs to change when one of those formats evolves. That is a legitimate separation-of-concerns win independent of LOC.

### Notes
- Author was explicit upfront about the smaller-than-projected LOC reduction (50 estimated → 9 actual) and surfaced an alternative larger seam (`SequencerTransitionApplier` around the `applyTransition` / `fireSequenceComplete` / `notifyStepChanged` cluster, ~60 LOC at lines 640-691). Worth queuing as a follow-up if further sequencer slimming is desired; not a reason to reject this PR.
- The adapter being stateless (chat cache threaded through, not held) is the right call — keeps the sequencer as the sole state owner and keeps the adapter trivially testable in isolation if dedicated tests are added later.

PR #525, opened 2026-05-18
refactor: extract ChatEventHandler from Plugin (#503)

cha-ndler:
## Summary

Extracts chat-message handling and the `/clh` chat command from `CollectionLogHelperPlugin` into a focused `com.collectionloghelper.chat.ChatEventHandler` collaborator, mirroring the LootSyncManager (#511) pattern.

- New `@Singleton` class `chat/ChatEventHandler.java` owns `handleChatMessage(ChatMessage)` and `onClhCommand(ChatMessage, String)`, plus the `COLLECTION_LOG_PATTERN` regex constant.
- Plugin keeps its `@Subscribe public void onChatMessage(ChatMessage)` method so the RuneLite event bus still discovers it via the existing `eventBus.register(plugin)` in `startUp`. The body is now a one-line delegation to `chatEventHandler.handleChatMessage(event)`.
- `chatCommandManager.registerCommand("clh", ...)` is rebound to `chatEventHandler::onClhCommand`.
- Two narrow callbacks (`rankedSourcesSupplier`, `onCollectionLogItemObtained`) bridge the handler back to the plugin's ranked-source cache so `getRankedSources()` and `rankedSourcesDirty` stay where the other cache flags live without crossing packages.
- Plugin.java drops the `COLLECTION_LOG_PATTERN` constant, the `onChatMessage` body, the `onClhCommand` method, and the `Matcher` / `Pattern` / `Text` / `ChatMessageType` / `CollectionLogItem` / `ClueCompletionEstimator` imports that were only used by the moved code.

Plugin.java: 1042 -> 959 LOC (-83). `@Inject` count: 18 -> 19 (+1 for the new handler). ChatEventHandler: 226 LOC.

Partially addresses #503.

## Test plan

- [x] `./gradlew build` succeeds (compile + checkstyle + tests + JaCoCo 45% coverage gate + lintDropRates).
- [x] Plugin.java still has `@Subscribe public void onChatMessage(ChatMessage event)` so the EventBus binding via `eventBus.register(plugin)` is preserved.
- [x] `/clh` command remains registered (now bound to `chatEventHandler::onClhCommand`) and unregistered in `shutDown`.
- [x] `COLLECTION_LOG_PATTERN` regex semantics unchanged (literal carried verbatim into the new class).
- [x] Cache invalidation parity verified: detecting a new collection-log item still sets `pendingPanelRebuild = true` and `rankedSourcesDirty = true` via the `onCollectionLogItemObtained` callback wired in `startUp`.
[...]

Issue #527, opened 2026-05-18
perf: avoid per-tick allocations in DynamicTargetManager guard-fail path (#523 follow-up)


cha-ndler:
## Background

PR #523 extracted `DynamicTargetManager` from `GuidanceOverlayCoordinator` as part of the #503 god-class breakdown. The extraction is correct, but it introduced a small per-tick allocation regression on guard-fail paths.

## Problem

`DynamicTargetManager.tick()` is invoked **every game tick** unconditionally. On guard-fail paths (i.e., when guidance is inactive or no dynamic target is in scope) the manager now allocates:

- 1 `new Input(...)` per tick
- 1 `new Result.unchanged(null)` per tick

On master prior to #523, those paths allocated nothing. This violates the project's overlay-performance discipline (per `feedback_overlay_performance.md`): hot-path code on every-tick handlers should be allocation-free on the common no-op path.

## Acceptance criteria

Either of the following mitigations is acceptable:

1. **Lift `isActive()` guard back to coordinator** so `DynamicTargetManager.tick()` is only invoked on active-guidance ticks. The coordinator already has the active-guidance check; calling the manager unconditionally moved the guard inside the manager and made the cheap-path allocate.
2. **Cache a singleton `Result.unchanged(null)` constant** on `DynamicTargetManager` and reuse a per-instance `Input` scratch object (mutating fields is acceptable since `Input` does not escape the method).

Tests should assert that the guard-fail path does not allocate (or at minimum that the manager is not invoked when guidance is inactive).

## Scope

Single PR, ~5-minute fix.

PR #528, opened 2026-05-18
refactor: extract TempleSyncOrchestrator from Plugin (#503)

cha-ndler:
## Summary

Wave 9 of the #503 god-class refactor. Extracts the TempleOSRS KC sync orchestration out of `CollectionLogHelperPlugin` into a focused collaborator in the `sync` package.

The new `TempleSyncOrchestrator` owns the end-to-end flow that the plugin's private `onTempleSyncRequested` method used to handle inline:

- Feature-toggle and player-name preconditions
- Submitting the sync to `TempleOsrsKcSyncer` and awaiting the result on the shared daemon executor (EDT/client thread stay unblocked)
- Filtering the returned KC map against the source database before writing to `SourceKcStore`
- Posting the user-facing chat message on the client thread and resetting the panel button

The plugin's panel-callback wiring becomes a one-line delegation; the `panel` and `httpResultExecutor` are passed per-call because their lifecycles are plugin-owned (matches the call-site pattern established by `LootSyncManager` in Wave 8).

## LOC delta

- `CollectionLogHelperPlugin.java`: 959 -> 866 LOC (-93)
- New: `sync/TempleSyncOrchestrator.java` (202 LOC, including license header + javadoc)
- New: test `sync/TempleSyncOrchestratorTest.java` (216 LOC)

## Test plan

- [x] New `TempleSyncOrchestratorTest` covers four observable paths:
  - feature toggle disabled (short-circuit)
  - missing / empty player name (panel notified, syncer never called)
  - successful result (KC store updated with DB-known sources only, unknown skipped)
  - failure result (KC store untouched, panel notified)
- [ ] Manual smoke: in-client, click `Sync KC from TempleOSRS` with sync enabled and a known character; verify chat message and button reset

## Related

- Tracking: #503
- Prior wave: #525 (ChatEventHandler extraction)

### DIFF
--- a/src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java
+++ b/src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java
@@ -34,8 +34,7 @@ import com.collectionloghelper.di.GuidanceModule;
 import com.collectionloghelper.di.SyncModule;
 import com.collectionloghelper.sync.ImportResult;
 import com.collectionloghelper.sync.LootSyncManager;
-import com.collectionloghelper.sync.SyncResult;
-import com.collectionloghelper.sync.TempleOsrsKcSyncer;
+import com.collectionloghelper.sync.TempleSyncOrchestrator;
 import com.collectionloghelper.efficiency.ScoredItem;
 import com.collectionloghelper.lifecycle.AuthoringLogger;
 import com.collectionloghelper.lifecycle.OverlayRegistry;
@@ -44,15 +43,12 @@ import com.collectionloghelper.lifecycle.SyncStateCoordinator;
 import com.collectionloghelper.ui.CollectionLogHelperPanel;
 import com.google.inject.Provides;
 import java.awt.image.BufferedImage;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
 import lombok.extern.slf4j.Slf4j;
@@ -148,6 +144,9 @@ public class CollectionLogHelperPlugin extends Plugin
 
 	@Inject
 	private LootSyncManager lootSyncManager;
+
+	@Inject
+	private TempleSyncOrchestrator templeSyncOrchestrator;
 
 	@Inject
 	private ChatEventHandler chatEventHandler;
@@ -256,7 +255,7 @@ public class CollectionLogHelperPlugin extends Plugin
 		});
 
 		// Wire TempleOSRS KC sync button callback
-		panel.setTempleSyncCallback(this::onTempleSyncRequested);
+		panel.setTempleSyncCallback(() -> templeSyncOrchestrator.requestSync(panel, httpResultExecutor));
 
 		// Wire coordinator with references it needs from the plugin
 		guidance.getGuidanceCoordinator().setPluginInstance(this);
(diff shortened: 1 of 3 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#528
**Files changed**: 3  (+423 / -98)
**Reviewer focus**: behavior-preservation of the extraction, hot-path allocation discipline on the guard-fail branches, executor/thread-safety vs. the `LootSyncManager` pattern (Wave 8), and test coverage of the four claimed observable paths.

### Findings

#### CRITICAL (0)
_None._

#### HIGH (0)
_None._

#### MEDIUM (1)

- **`SYNC_TIMEOUT_SECONDS = 30` is duplicated across sync collaborators**. The 30s wait-for-future was inlined in the plugin and is now inlined in `TempleSyncOrchestrator`. If `LootSyncManager` (or a future syncer) ever wants the same value, consider lifting the timeout onto `TempleOsrsKcSyncer` (the producer of the `Future`) or onto a shared `SyncConstants`. Not blocking — fine to defer to a future wave.

#### LOW (3)

- The two guard branches (`!config.enableTempleOsrsSync()` and missing-player-name) behave inconsistently w.r.t. the panel: the disabled-feature branch does **not** call `panel.onTempleSyncComplete(false)`, while the missing-name branch does. This matches the pre-refactor behavior exactly, so it's a behavior-preserving extraction — but worth a one-line comment explaining why the disabled branch doesn't need to reset the button (presumably because the button is hidden/disabled when the feature is off). Optional polish.
- `requestSync(panel, resultExecutor)` accepts a `@Nullable`-in-spirit panel (javadoc says so) but a non-annotated `ExecutorService`. If `resultExecutor` is ever null during shutdown races, the NPE will surface on the callback thread. Either annotate or add a null-check at the top to mirror the panel handling.
- The success-path `HashMap<String, Integer> validated = new HashMap<>()` is allocated per-call, but only on the success branch — that's correct behavior. No allocation on the disabled-feature guard fail, which is the hot path concern from the #527/#523 lesson. Confirmed clean.

### Verification ran

- [x] `./gradlew test --tests TempleSyncOrchestratorTest` — 5/5 passed (4 named paths + the success path is implicitly two cases via null/empty being split)
- [x] CI status checks (build, CI) — both SUCCESS
- [x] Mergeable: CLEAN
- [x] Diff inspection: extraction is line-for-line behavior-preserving; the plugin's `onTempleSyncRequested` body maps 1:1 onto `requestSync` + `awaitResult` + `applySuccessfulResult` + `postChatMessage`. The only observable difference is the helper-method split, which is desired.
- [x] Executor/thread discipline: `httpResultExecutor` is created in `startUp` and `shutdownNow()` in `shutDown` (plugin lines 193/355–358). The orchestrator borrows it per-call rather than holding a reference — same pattern as `LootSyncManager` (`@Singleton` + executor passed at the call site). Client-thread side-effects routed through `clientThread.invokeLater` exactly as before.
- [x] Hot-path allocation check: feature-disabled guard returns before any allocation. Missing-name guard allocates only the `WARN` log args (unavoidable for the diagnostic). No per-tick path involved — this is a button callback.
- [x] No new `client.getWidget(WidgetInfo.X)`, no `Player.getPolygons()`, no overlay render-path changes.

### Recommendation

**SHIP** — no CRITICAL or HIGH issues. The extraction is a clean, behavior-preserving 93-LOC reduction in the god-class, matches the Wave 8 `LootSyncManager` call-site pattern, has meaningful test coverage of all four observable paths (disabled / missing-name / success-with-DB-filter / failure), CI is green, and the merge state is CLEAN.

### Notes

- Plugin LOC trend after this wave: 959 → 866. Combined with prior waves (ChatEventHandler #525, etc.), the #503 god-class breakdown is making steady, measurable progress.
- The test class deliberately constructs a real `CollectionLogSource` rather than mocking it because the type is Lombok `@Value` (final). The inline comment explains this clearly — good defensive documentation against future Mockito-upgrade confusion.
- Suggested follow-up (not blocking): the next sync collaborator extraction could consolidate `SYNC_TIMEOUT_SECONDS` into a single location. Flagging for the orchestrator to consider sequencing.

Issue #429, opened 2026-05-14
bug: Minimap guidance arrow no longer clickable to focus worldmap (regression)


cha-ndler:
The minimap guidance arrow used to be click-through: hovering or clicking it would focus the worldmap on the arrow's target. That behavior is gone.

Likely a regression introduced by recent overlay work — candidates: PR #393 (GuidanceMovementTracker / hint arrow refresh) or PR #397 (WorldMapDestinationOverlay arrow + icon).

Reproducer: activate any guidance with a worldmap target (e.g. Mort'ton, Vorkath); hover/click the minimap arrow indicator; expected: worldmap focuses on target tile; observed: nothing.

Severity: MEDIUM (regression of working behavior). Surfaced 2026-05-13 in-client validation pass (F2).

Issue #410, opened 2026-05-13
bug: world map shows two overlapping edge arrows when guidance destination is off-screen


cha-ndler:
## Summary

When guidance is active and the destination tile is off-screen on the world map, **two edge-snap arrows render simultaneously** at the map boundary, both pointing toward the destination.

This is the issue flagged by the [B.5.5 worker's PR #397 body](https://github.com/cha-ndler/collection-log-helper/pull/397) as a known product decision deferred for in-client validation.

## Source

The two arrows come from independent overlay systems:
- **`WorldMapDestinationOverlay`** (new, from PR #397 / B.5.5) — the per-step edge-snap arrow + destination icon
- **`CollectionLogWorldMapPoint`** — the legacy worldmap chevron managed via `WorldMapPointManager` in `GuidanceOverlayCoordinator`

Both render in teal, so they're visually consistent, but the doubled chevron is cluttered.

## Evidence

- Surfaced during 2026-05-13 in-client validation pass (T3) against master HEAD `c2cdd81e`
- Operator quote: "edge arrow at boundary, but it looks strange, like two arrows overlapping"
- Screenshot: \`C:\Users\Chandler\AppData\Local\Temp\runelite_screenshot_1778679630348.png\` (operator-local; do not commit)

## Suggested fix (pick one)

**Option A**: Hide `CollectionLogWorldMapPoint` while CLH guidance is active. The legacy chevron is superseded by the new overlay's arrow + icon. `WorldMapPointManager.remove(activeMapPoint)` on activation, restore in `deactivateGuidance`.

**Option B**: Remove the new `WorldMapDestinationOverlay` edge-snap arrow when a `CollectionLogWorldMapPoint` is already drawn (off-screen case only). Keep the new overlay's icon for the on-screen case.

Option A is simpler and aligns with PR #397's intent (the new overlay supersedes the legacy chevron). Recommended.

## Scope

~30 LOC. Touches `GuidanceOverlayCoordinator` activation/deactivation paths. No schema or data change. Existing tests should still pass.

## Severity

LOW (cosmetic only — functionality is correct, just visually doubled).

## Refs

- Surfaces: PR #397 (B.5.5) acknowledged the overlap pre-merge.
- Discovered: 2026-05-13 in-client validation pass.

PR #529, opened 2026-05-18
refactor: extract OverlaySourceApplier from GuidanceOverlayCoordinator (#503)

cha-ndler:
## Summary

Wave 9 of #503: extract the `applySourceToOverlays` cluster from `GuidanceOverlayCoordinator` into a new `OverlaySourceApplier` Guice singleton. Pure structural extraction — no behavioral change.

Follows the established pattern from Wave 7 (`OverlayStepApplier`):
- `@Singleton @Inject` Guice component
- Holds no mutable state; coordinator-owned state (`activeMapPoint`, `pendingShortestPathTarget`) is threaded through `Input`/`Result` carriers
- Hint-arrow predicate and shortest-path setter passed as functional callbacks so the applier does not take a hard dependency on `WorldMapController` or the coordinator field

## LOC delta

- `GuidanceOverlayCoordinator.java`: 740 -> 711 (-29)
- `OverlaySourceApplier.java`: new (209 LOC, copyright header + 11-param ctor + Input/Result carriers)
- Coordinator still above the 600-LOC target — additional waves will continue.

## Test plan

- [x] `./gradlew test` passes (all existing tests, including 4 reflection-based coordinator tests updated to inject the new singleton)
- [x] Public API of `GuidanceOverlayCoordinator` unchanged
- [x] `applySourceToOverlays` body unchanged — only relocated; same overlay setter sequence, same `clientThread.invokeLater` semantics for hint arrow + ShortestPath

### DIFF
--- a/src/main/java/com/collectionloghelper/guidance/GuidanceOverlayCoordinator.java
+++ b/src/main/java/com/collectionloghelper/guidance/GuidanceOverlayCoordinator.java
@@ -112,6 +112,7 @@ public class GuidanceOverlayCoordinator
 	private final GroundItemHighlightOverlay groundItemHighlightOverlay;
 	private final WidgetHighlightOverlay widgetHighlightOverlay;
 	private final OverlayStepApplier overlayStepApplier;
+	private final OverlaySourceApplier overlaySourceApplier;
 	private final WorldMapController worldMapController;
 	private final DynamicTargetManager dynamicTargetManager;
 	private final NpcTrackerHelper npcTrackerHelper;
@@ -182,6 +183,7 @@ public class GuidanceOverlayCoordinator
 		GroundItemHighlightOverlay groundItemHighlightOverlay,
 		WidgetHighlightOverlay widgetHighlightOverlay,
 		OverlayStepApplier overlayStepApplier,
+		OverlaySourceApplier overlaySourceApplier,
 		WorldMapController worldMapController,
 		DynamicTargetManager dynamicTargetManager,
 		NpcTrackerHelper npcTrackerHelper)
@@ -208,6 +210,7 @@ public class GuidanceOverlayCoordinator
 		this.groundItemHighlightOverlay = groundItemHighlightOverlay;
 		this.widgetHighlightOverlay = widgetHighlightOverlay;
 		this.overlayStepApplier = overlayStepApplier;
+		this.overlaySourceApplier = overlaySourceApplier;
 		this.worldMapController = worldMapController;
 		this.dynamicTargetManager = dynamicTargetManager;
 		this.npcTrackerHelper = npcTrackerHelper;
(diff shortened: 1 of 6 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#529
**Files changed**: 6 (+295/-47)
**Reviewer focus**: Hot-path allocation discipline on the Input/Result carriers, structural parity with `OverlayStepApplier` (Wave 7), and mechanical correctness of the 4 reflection-based coordinator test updates.

### Findings

#### CRITICAL (0)
- None.

#### HIGH (0)
- None.

#### MEDIUM (1)
- `OverlaySourceApplier.java` ships with `@Slf4j` but no `log.*` call sites inside the class. Minor; either drop the annotation or add a debug log on apply (`OverlayStepApplier` is also `@Slf4j` so keeping it for parity is defensible). Not a blocker.

#### LOW (2)
- The `final CollectionLogWorldMapPoint mapPointForReturn = updatedMapPoint;` re-binding on line 156 is only needed because `updatedMapPoint` is reassigned earlier; a single `final` reassignable pattern (`CollectionLogWorldMapPoint newMapPoint = new CollectionLogWorldMapPoint(...)` after the conditional remove) would let you drop the alias. Cosmetic.
- `Input` carries `collectionLogIcon` as a separate field rather than reading it from a coordinator-injected field, which is fine for now but means callers must remember to thread it through. If a future wave moves `collectionLogIcon` ownership, this becomes redundant. Tracking-only.

### Hot-path / allocation analysis

`applySourceToOverlays` has exactly one call site (`GuidanceOverlayCoordinator.java:349`) inside `activateGuidance`, which fires on user-initiated guidance activation — not per tick, per varbit, or per overlay render. The two new short-lived allocations per call (`Input`, `Result`) are functionally amortised to once per activation. This does **not** repeat the #523/#527 per-tick allocation pattern. No concern.

### Pattern consistency

Mirrors `OverlayStepApplier` (Wave 7) precisely:
- Same `@Slf4j @Singleton` annotations.
- Same package (`com.collectionloghelper.guidance`) — matches siblings (`OverlayStepApplier`, `DynamicTargetManager`, `NpcTrackerHelper`, `WorldMapController`). Agree with Psi's call: a `guidance.overlay` subpackage would split the family for no clear win at this stage. If the guidance package later grows past ~15 classes, that re-org can be its own PR.
- Same `Input` / `Result` static final nested-class shape, package-private constructor, `@Nullable` fields.
- Functional callbacks for `hintArrowPredicate` and `pendingShortestPathTargetSetter` keep the applier from taking a hard dependency on `WorldMapController` or the coordinator field — clean.

### Behavior preservation

Diffed the moved body against the deleted block. Verbatim match on:
- Overlay setter sequence (`guidanceOverlay.set*`, `dialogHighlightOverlay.set*`, `guidanceMinimapOverlay.setTargetPoint`, `worldMapRouteOverlay.setTargetPoint`, `worldMapDestinationOverlay.setTarget` with `TILE` icon).
- World map point lifecycle: remove-existing-then-replace, `setMapPointActive(true)`.
- `clientThread.invokeLater` body: `clearHintArrow` -> conditional `setHintArrow` -> `useShortestPath` branch with `PluginMessage("shortestpath", "clear")` -> assign pending target.
- Comment annotations referencing #429 and #410 retained.

The only semantic delta is that `pendingShortestPathTarget` assignment now goes through a `Consumer<WorldPoint>` callback rather than direct field write. Both run on the client thread (inside `invokeLater`), so the deferred-write ordering is preserved.

### Test-update review

All 4 reflection-based coordinator tests (`DynamicTargetEvaluatorDispatchTest`, `GuidanceOverlayCoordinatorDescriptionTest`, `GuidanceOverlayCoordinatorMapPointTest`, `GuidanceOverlayCoordinatorNpcIdTest`) updated identically:
1. New `newOverlaySourceApplier()` helper added to each (duplicated across files — see Note below).
2. Constructor signature lookup adds `OverlaySourceApplier.class` in the correct positional slot (after `OverlayStepApplier.class`, before `WorldMapController.class`), matching the coordinator's parameter order.
3. `ctor.newInstance(...)` call sites updated to pass the new singleton in the matching slot.

Slot ordering is consistent across all four tests and matches the production constructor. No drift.

### Verification ran
- [x] `./gradlew test --rerun-tasks` — 1205 tests, 0 failures, 0 errors (JDK 17).
- [x] Build successful.
- [x] Coordinator LOC: 740 -> 711 (-29). Still above the 600-LOC target — additional waves will continue.
- [x] `OverlaySourceApplier` 209 LOC, well under 800.
- [x] `applySourceToOverlays` call-site count verified (1 call site, non-hot-path).

### Recommendation
**SHIP** — no CRITICAL or HIGH issues. Clean structural extraction with behavior preserved verbatim, tests updated mechanically, build green. The MEDIUM `@Slf4j`-with-no-logs note is parity with `OverlayStepApplier` and not worth blocking on.

### Notes
- **Duplication across tests**: each of the 4 coordinator tests now has its own `newOverlaySourceApplier()` helper (mirroring the same pattern for `newOverlayStepApplier()`). When the next wave lands, a small test-fixtures helper (`CoordinatorTestFixtures` or similar) could DRY up the constructor-by-reflection ceremony. Out of scope for this PR; flagging for a future cleanup pass.
- **Coordinator still above 600 LOC target** (711). Per the PR body, more waves to follow. Recommend tracking remaining extractable clusters in #503 so the wave plan stays explicit.
- **Suggestion for the checklist**: add a "delegate-callback vs. injected-dep" guideline — Psi correctly avoided a hard dep on `WorldMapController` here by passing a `Predicate`. Worth codifying as a Wave-N pattern to apply consistently.

PR #515, opened 2026-05-18
refactor: extract SyncButtonController from Panel (#503)

cha-ndler:
## Summary

Pulls the two sync buttons out of `CollectionLogHelperPanel` into a dedicated `SyncButtonController` helper. Same construction pattern as `PanelRebuildOrchestrator` (#513): not a Guice singleton, instantiated by the panel constructor, bound to the panel's Swing tree.

The new controller owns:

- The "Sync from collectionlog.net" and "Sync KC from TempleOSRS" `JButton` instances (labels, fonts, tooltips, sizes, action listeners)
- The two `Runnable` click callbacks wired by the plugin after panel construction
- Per-config visibility refreshes (`updateCollectionLogNetImportButton`, `updateTempleSyncButtonVisibility`)
- Post-sync enabled/text reset (`onCollectionLogNetImportComplete`, `onTempleSyncComplete`)

Panel changes:

- Six public sync-button methods become one-line delegations; their signatures are unchanged so external callers (`CollectionLogHelperPlugin`, `SyncStateCoordinator`, `LootSyncManager`) are unaffected.
- Removed fields: `collectionLogNetSyncButton`, `templeSyncButton`, `collectionLogNetImportCallback`, `templeSyncCallback`.
- Removed now-unused `javax.swing.JButton` import.

EDT discipline preserved: methods called off the EDT keep their `SwingUtilities.invokeLater` wrappers inside the controller; methods invoked only from action listeners mutate components directly.

## Metrics

- Panel LOC: **861 -> 800** (now at the issue #503 target).
- Controller LOC: 218 (new).
- Test LOC: 235 (13 new tests).
- Build: green. Tests: 1195 passing, 0 failures. JaCoCo 45% line-coverage gate holds.

Partially addresses #503.

## Test plan

- [x] `./gradlew compileJava` succeeds
- [x] `./gradlew test --tests com.collectionloghelper.ui.SyncButtonControllerTest` — 13 tests pass
- [x] `./gradlew build` succeeds (includes JaCoCo 45% gate)
- [x] Full test suite: 1195 tests, 0 failures, 0 errors
- [x] Panel LOC verified at 800
- [ ] Manual verification: toggle `enableCollectionLogNetImport` and `enableTempleOsrsSync` in config — buttons show/hide; click each button and confirm sync runs and label/enabled state restores on completion
- [ ] Manual verification: panel layout order in controls strip is unchanged (collectionlog.net button still appears above the TempleOSRS button)

PR #531, opened 2026-05-18
refactor: extract SelectorControlsPanel from Panel (#503)

cha-ndler:
## Summary

Wave 10 of the #503 god-class split. Moves the five selector widgets (mode, AFK-filter, sort-mode, category, search) plus their listeners, visibility rules, and the search debounce timer out of `CollectionLogHelperPanel` into a dedicated `SelectorControlsPanel`.

- Panel keeps authority over `currentMode`; the selector panel reports changes via callbacks supplied at construction.
- Sibling of `PanelRebuildOrchestrator`, `SyncButtonController`, `ListContainerScrollPane`, `DetailViewBuilder`. Panel-constructed (not @Singleton), bound to panel lifecycle.
- `CollectionLogHelperPanel.java`: 763 -> 685 LOC (-78).
- New `SelectorControlsPanel.java`: 228 LOC.
- New `SelectorControlsPanelTest.java`: 9 unit tests covering initial visibility, visibility transitions, callback invocations, query normalization, and programmatic mode selection.

## Test plan

- [x] `./gradlew compileJava` clean
- [x] `./gradlew test` passes (1219 tests)
- [x] New `SelectorControlsPanelTest` covers visibility rules per mode, all five change callbacks, search-query trimming/lower-casing, and programmatic selection
- [x] Public API of `CollectionLogHelperPanel` unchanged (`setMode`, `getSearchQuery`, `getEfficientSortMode`, `getSelectedCategory`, `switchToCategoryFocus`, `shutDown` preserved)

### DIFF
--- a/src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java
+++ b/src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java
@@ -323,10 +323,21 @@ public class CollectionLogHelperPanel extends PluginPanel implements PanelShellC
 
 	public void setMode(Mode mode)
 	{
-		currentMode = mode;
+		// Route programmatic mode changes through the selector listener so the
+		// mode-controller lifecycle hooks (onModeDeactivated / onModeActivated)
+		// fire just like a user-driven selection. Mutating currentMode here
+		// before calling setSelectedMode(...) would cause the listener to see
+		// previous == next and short-circuit the dispatcher, silently breaking
+		// the contract for any controller that overrides those hooks.
+		if (mode == currentMode)
+		{
+			// Same-mode call: the combo listener won't fire, so apply the
+			// visibility + rebuild side effects directly.
+			updateControlVisibility();
+			rebuild();
+			return;
+		}
 		selectorControls.setSelectedMode(mode);
-		updateControlVisibility();
-		rebuild();
 	}
 
 	private void onModeSelected(Mode mode)
--- a/src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java
+++ b/src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java
@@ -67,24 +67,18 @@ import java.awt.CardLayout;
 import java.awt.Color;
 import java.awt.Component;
 import java.awt.Dimension;
-import java.awt.event.ItemEvent;
 import java.util.List;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import javax.swing.BorderFactory;
 import javax.swing.Box;
 import javax.swing.BoxLayout;
-import javax.swing.JComboBox;
 import javax.swing.JLabel;
 import javax.swing.JPanel;
 import javax.swing.JProgressBar;
 import javax.swing.JScrollPane;
-import javax.swing.JTextField;
 import javax.swing.SwingConstants;
 import javax.swing.SwingUtilities;
-import javax.swing.Timer;
-import javax.swing.event.DocumentEvent;
-import javax.swing.event.DocumentListener;
 import net.runelite.client.game.ItemManager;
 import net.runelite.client.ui.ColorScheme;
 import net.runelite.client.ui.FontManager;
@@ -142,11 +136,7 @@ public class CollectionLogHelperPanel extends PluginPanel implements PanelShellC
 	private final ClueSummaryView clueSummaryView;
 	private final SyncButtonController syncButtonController;
 
-	private final JComboBox<Mode> modeSelector;
-	private final JComboBox<AfkFilter> afkFilterSelector;
-	private final JComboBox<EfficientSortMode> sortSelector;
-	private final JComboBox<CollectionLogCategory> categorySelector;
-	private final JTextField searchField;
+	private final SelectorControlsPanel selectorControls;
 	private final JLabel completionLabel;
 	private final JProgressBar completionProgressBar;
 	private final JPanel listContainer;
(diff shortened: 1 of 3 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#531
**Files changed**: 3 (+529/-110)
**Reviewer focus**: Callback discipline / `currentMode` authority, listener-rebuild semantics, debounce timer lifecycle, test hygiene, behavior parity with pre-extraction inline listeners.

### Findings

#### CRITICAL (0)
_None._

#### HIGH (1)

- **Programmatic `setMode(...)` no longer fires mode-lifecycle hooks (latent regression).**
  In `CollectionLogHelperPanel.setMode(mode)`, the new ordering is:
  ```java
  currentMode = mode;                       // mutate first
  selectorControls.setSelectedMode(mode);   // triggers ItemListener -> onModeSelected
  ```
  Inside `onModeSelected`, `Mode previous = currentMode;` captures the **already-updated** value, so `modeDispatcher.switchMode(previous, currentMode)` is called with `(newMode, newMode)` and the dispatcher's `previous == next` early-return skips `onModeDeactivated()` / `onModeActivated()`.

  Pre-extraction, the inline listener ran first (because `modeSelector.setSelectedItem(mode)` triggered it before `setMode` mutated `currentMode`), so the lifecycle hooks fired correctly.

  Today every controller uses the default no-op hooks (no overrides found), so this is **observable-but-harmless in current code** — but it silently breaks the contract for any future controller that overrides `onModeActivated`/`onModeDeactivated`. The only programmatic caller is `CollectionLogHelperPlugin.startUp() -> panel.setMode(config.defaultMode())`, so users whose `defaultMode` is not `EFFICIENT` will lose the activation hook on plugin start once any controller starts using it.

  Suggested fix: move the `currentMode = mode` line **after** `selectorControls.setSelectedMode(mode)`, or refactor `setMode` to delegate entirely to the listener path (drop the explicit `currentMode = mode` and the trailing `updateControlVisibility()` / `rebuild()` so it just calls `selectorControls.setSelectedMode(mode)` and lets `onModeSelected` do all the work).

#### MEDIUM (1)

- **`onSearchOrFilterChanged` is overloaded for two unrelated triggers.** The category combo's listener and the search-field debounce both fire the same `Runnable`. That preserves prior behavior (both called `rebuild()`), but it makes the constructor signature slightly misleading — a renamed `onRebuildRequested` or separate `onCategoryChanged` / `onSearchTextChanged` callbacks would be clearer and let future code differentiate between "user typed" vs "user picked a category". Not a blocker; flagging for future API hygiene.

#### LOW (2)

- **Initial visibility duplicates the post-construction `updateControlVisibility()` call.** `SelectorControlsPanel`'s ctor seeds `afkFilterSelector`/`sortSelector`/`categorySelector`/`searchField` visibility, then the panel calls `updateControlVisibility()` at end of its own ctor (line 316), which immediately overwrites those seeds. Behavior matches pre-PR (which had the same redundancy), but the new class could be simplified by setting all four to `false` initially and trusting the caller's `updateVisibility(initialMode)`.
- **`SelectorControlsPanelTest.findCombo` empty-combo edge case.** The helper requires `combo.getItemCount() > 0` and uses `itemType.isInstance(combo.getItemAt(0))` to identify the right combo. Works today (all four enums have values), but if `CollectionLogCategory` ever ships with zero values the helper would silently skip and throw — a small defensive comment or a tag-based lookup would be more robust.

### Verification ran
- [x] `./gradlew test` locally — 1219 tests pass on Windows JDK-17
- [x] New `SelectorControlsPanelTest` (9 tests) — all green; runs everything on EDT via `SwingUtilities.invokeAndWait`, no scrollbar mutation, no flake risk
- [x] `PanelRebuildOrchestratorTest` passes locally; `pinScrollBounds(...)` already applied to the failing test (line 154)
- [x] CI failure is the known JDK-17 / headless-Linux `PanelRebuildOrchestratorTest.captureRecordsScrollPositionFromEnclosingScrollPane` flake (#515/#516 lineage); unrelated to this PR — recommend `--rerun-tasks` once the HIGH finding is addressed
- [x] Public API of `CollectionLogHelperPanel` preserved (`setMode`, `getSearchQuery`, `getEfficientSortMode`, `getSelectedCategory`, `switchToCategoryFocus`, `shutDown`)
- [x] Debounce timer is stopped via `selectorControls.shutDown()` in panel's `shutDown()` — no timer leak
- [x] File sizes: `SelectorControlsPanel` 228 LOC, panel reduced 763 → 685 LOC, test 269 LOC — all under 800 LOC limit
- [x] No new `getWidget(WidgetInfo...)` / `Player.getPolygons()` usage, no overlay hot-path allocations (not an overlay change)

### Recommendation
**NEEDS WORK** — one HIGH finding worth addressing: re-order the `setMode` body (or rework it to delegate to the listener) so `modeDispatcher.switchMode(previous, next)` receives the actual prior mode. Tiny diff, restores pre-extraction contract, prevents a latent footgun for future controllers.

CI flake is unrelated (known scrollbar issue); not a blocker. MEDIUM and LOW items are polish.

### Notes
- The `switchToCategoryFocus` path is **not** affected: it calls `selectorControls.setSelectedMode(...)` directly without pre-mutating `currentMode`, so the listener captures the correct `previous` mode and dispatcher fires normally. Only the programmatic `setMode(...)` entry point has the regression.
- Consider adding a unit test that exercises programmatic `setMode(...)` through the panel and asserts the dispatcher's previous/next args — this would have caught the regression. The existing `SelectorControlsPanelTest.setSelectedMode_updatesCombo` only checks the combo state, not the lifecycle hooks fired on the panel side.
- Suggested checklist add: for callback-driven extractions of inline `ItemListener` code, verify the order of state mutation vs `setSelectedItem(...)` in any programmatic setters that drive the new widget — pre/post-mutation flips listener-captured `previous` values silently.

PR #532, opened 2026-05-18
refactor: extract OverlayDeactivator from GuidanceOverlayCoordinator (#503)

cha-ndler:
## Summary

Wave 10 of the GuidanceOverlayCoordinator decomposition (#503). Extracts the overlay deactivation cluster into a new `OverlayDeactivator` helper that follows the Input/Result carrier pattern established by `OverlaySourceApplier` (#529).

- `clearGuidanceOverlays()` (partial between-step clear) and `deactivateGuidance()` (full teardown) now delegate to `OverlayDeactivator.clearOverlays(Input)` and `OverlayDeactivator.deactivate(Input)` respectively.
- Coordinator-owned mutable state (`activeMapPoint`, `activeInfoBox`, `pendingShortestPathTarget`) flows in via `Input` and out via `Result`, so the coordinator remains the single owner of that state.
- Pure extraction. Behaviour preserved exactly (same overlay calls, same order, same client-thread deferral, same ShortestPath plugin-message dispatch).

## LOC delta

`GuidanceOverlayCoordinator.java`: **711 -> 667 (-44 LOC)**.

Closer to the 600-LOC target. Remaining seam highlighted as Wave 11 candidate: the `onStepChanged` cluster (~60 LOC at the bottom of the file), which is genuinely stateful and will need careful Input/Result discipline.

## Test plan

- [x] `./gradlew test` passes (covered by `build`)
- [x] Reflection-based coordinator tests updated to inject the new singleton through the package-private constructor:
  - [x] `GuidanceOverlayCoordinatorMapPointTest`
  - [x] `GuidanceOverlayCoordinatorNpcIdTest`
  - [x] `GuidanceOverlayCoordinatorDescriptionTest`
  - [x] `DynamicTargetEvaluatorDispatchTest`
- [ ] Manual smoke: start guidance for a multi-step source, step through, then stop guidance; verify overlays clear, hint arrow clears, ShortestPath "clear" fires, and panel resets.
- [ ] Manual smoke: start guidance for a single-step source and stop guidance; verify the InfoBox is removed and `activeMapPoint` is unregistered from `WorldMapPointManager`.

### DIFF
--- a/src/main/java/com/collectionloghelper/guidance/GuidanceOverlayCoordinator.java
+++ b/src/main/java/com/collectionloghelper/guidance/GuidanceOverlayCoordinator.java
@@ -113,6 +113,7 @@ public class GuidanceOverlayCoordinator
 	private final WidgetHighlightOverlay widgetHighlightOverlay;
 	private final OverlayStepApplier overlayStepApplier;
 	private final OverlaySourceApplier overlaySourceApplier;
+	private final OverlayDeactivator overlayDeactivator;
 	private final WorldMapController worldMapController;
 	private final DynamicTargetManager dynamicTargetManager;
 	private final NpcTrackerHelper npcTrackerHelper;
@@ -184,6 +185,7 @@ public class GuidanceOverlayCoordinator
 		WidgetHighlightOverlay widgetHighlightOverlay,
 		OverlayStepApplier overlayStepApplier,
 		OverlaySourceApplier overlaySourceApplier,
+		OverlayDeactivator overlayDeactivator,
 		WorldMapController worldMapController,
 		DynamicTargetManager dynamicTargetManager,
 		NpcTrackerHelper npcTrackerHelper)
@@ -211,6 +213,7 @@ public class GuidanceOverlayCoordinator
 		this.widgetHighlightOverlay = widgetHighlightOverlay;
 		this.overlayStepApplier = overlayStepApplier;
 		this.overlaySourceApplier = overlaySourceApplier;
+		this.overlayDeactivator = overlayDeactivator;
 		this.worldMapController = worldMapController;
 		this.dynamicTargetManager = dynamicTargetManager;
 		this.npcTrackerHelper = npcTrackerHelper;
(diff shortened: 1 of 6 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#532
**Files changed**: 6 (+370 / -69)
**Reviewer focus**: dependency direction (OverlayDeactivator -> GuidanceSequencer), behavior parity vs old inline clearGuidanceOverlays/deactivateGuidance, hot-path discipline, reflection-test consistency with Wave-9 pattern.

### Findings

#### CRITICAL (0)
None.

#### HIGH (0)
None.

#### MEDIUM (1)

- **Coordinator still above 600-LOC target (667 LOC).** Acknowledged in the PR body as a Wave 11 candidate (`onStepChanged` cluster, ~60 LOC). Not a blocker; flagging only to keep the runway visible.

#### LOW (2)

- `OverlayDeactivator.deactivate` reassigns the local `infoBox` to `null` after `infoBoxManager.removeInfoBox(infoBox)`. The reassignment is dead (we immediately return `new Result(null, infoBox, null)`). Harmless, but you could drop the `infoBox = null;` line for clarity. Keeping it does mirror the old field-write structurally, which is arguably what you want for line-by-line parity reviewability — fine to leave.
- The two reflection-test helpers (`newOverlayDeactivator`) are duplicated verbatim across four test classes. Consider extracting to a shared test fixture in a follow-up (similar shape to the existing `newOverlaySourceApplier` duplication). Not new debt — just an observation that this PR inherits the existing Wave-9 pattern.

### Verification ran

- [x] `./gradlew test --tests *OverlayDeactivator* --tests *GuidanceOverlayCoordinator* --tests *DynamicTargetEvaluatorDispatch*` — all green locally on JDK 17.
- [x] CI workflow (`CI / build`) green on this branch.
- [x] Failing `Build / build` check is the known-flaky `PanelRebuildOrchestratorTest > captureRecordsScrollPositionFromEnclosingScrollPane` JDK-11 scrollbar flake (unrelated to this PR — same flake hit several recent PRs in this wave).
- [x] Dependency direction verified: `OverlayDeactivator` injects `GuidanceSequencer` only to call `stopSequence()` on full teardown. `GuidanceSequencer` does not inject `OverlayDeactivator` or `GuidanceOverlayCoordinator` — no cycle. The direction is justified: deactivator is the orchestration boundary calling into the sequencer to halt, which matches the pre-extraction call graph (the coordinator called `guidanceSequencer.stopSequence()` from the same method body).
- [x] Behavior parity verified line-by-line vs `deactivateGuidance`/`clearGuidanceOverlays` on `master`:
  - `deactivate`: same order — `npcTrackerHelper.clear()` -> InfoBox remove -> `guidanceSequencer.stopSequence()` -> overlay clears (guidance, minimap, route, destination + setMapPointActive(false), dialog, object, item, groundItem, widget) -> `worldMapPointManager.removeIf(...)` -> `clientThread.invokeLater(clearHintArrow + optional shortestpath PluginMessage)` -> panel hideClueGuidance/hideStepProgress/setGuidanceState -> log.debug. The old field nulls for `activeMapPoint` and `pendingShortestPathTarget` are now applied at the coordinator via `Result` write-back; they happen after the synchronous clears and before the lambda is dispatched on the client thread — same observable timing.
  - `clearOverlays`: same partial-clear set; `activeInfoBox` and `pendingShortestPathTarget` correctly passed through unchanged via `Result`. Only `activeMapPoint` is nulled, matching old behaviour.
- [x] Hot-path discipline: neither `clearOverlays` nor `deactivate` is invoked from `onGameTick`/`onClientTick`/render. Call sites in the coordinator are user-stop and step-transition. No new per-tick allocation introduced. (Cross-checked against #527/#523 lessons.)
- [x] Reflection-test pattern matches Wave-9 (`newOverlaySourceApplier`): package-private constructor, `setAccessible(true)`, same arg ordering as DI signature. All four test classes consistent.

### Recommendation
**SHIP** — no CRITICAL or HIGH issues. Pure extraction; behavior parity preserved; tests pass; dependency direction is sound (no cycle, justified by call graph); not on a per-tick path. The only failing check is a pre-existing unrelated test flake.

### Notes

- The Input/Result carrier pattern continues to scale well. Wave 11 (the `onStepChanged` cluster called out in the PR body) is genuinely stateful — recommend explicit `Input`/`Result` discipline there too so the coordinator stays the single owner of state.
- Worth tracking in MEMORY: across Waves 9 and 10, four reflection-based coordinator tests now each carry near-duplicate factory helpers. A shared `CoordinatorTestFixture` would pay back as Wave 11 onward grow the constructor further.

PR #533, opened 2026-05-18
refactor: extract CollectionLogNetImportOrchestrator from Plugin (#503)

cha-ndler:
## Summary

Wave 11 of the #503 god-class breakdown. Extracts the collectionlog.net import callback wiring (player-name lookup + async dispatch + result handoff) from `CollectionLogHelperPlugin.startUp()` into a dedicated `@Singleton` orchestrator in `com.collectionloghelper.sync`.

Mirrors the Wave 9 `TempleSyncOrchestrator` pattern: the panel and the shared http-result executor are passed per-call so the plugin retains lifecycle ownership.

### LOC delta

- `CollectionLogHelperPlugin.java`: **780 -> 753** (-27)
- New: `CollectionLogNetImportOrchestrator.java` (~130 LOC incl. license + javadoc)
- New: `CollectionLogNetImportOrchestratorTest.java` (5 paths)

### Behaviour

No behavioural change. The new class re-creates the exact callback that previously lived inline:

1. Resolve `client.getLocalPlayer().getName()`; if null/empty toast `"Log in first"` and return.
2. Submit `CollectionLogNetImporter.importProfile(username)`.
3. On the supplied executor, await the future and forward the toast message to the panel.
4. If `result.isSuccess()`, call `panel.rebuild()`.
5. On any exception, log and surface `"collectionlog.net: error"`.

## Test plan

- [x] `./gradlew test` passes
- [x] New `CollectionLogNetImportOrchestratorTest` covers: no local player, empty player name, success (rebuild triggered), non-success (no rebuild), future failure (generic error message)
- [ ] Manual smoke: click "Import from collectionlog.net" in the panel while logged in and confirm toast + obtained-item state updates
- [ ] Manual smoke: click the same button while logged out and confirm the "Log in first" toast

### DIFF
--- a/src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java
+++ b/src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java
@@ -32,7 +32,7 @@ import com.collectionloghelper.di.DataModule;
 import com.collectionloghelper.di.EfficiencyModule;
 import com.collectionloghelper.di.GuidanceModule;
 import com.collectionloghelper.di.SyncModule;
-import com.collectionloghelper.sync.ImportResult;
+import com.collectionloghelper.sync.CollectionLogNetImportOrchestrator;
 import com.collectionloghelper.sync.LootSyncManager;
 import com.collectionloghelper.sync.TempleSyncOrchestrator;
 import com.collectionloghelper.efficiency.ScoredItem;
@@ -49,7 +49,6 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
 import lombok.extern.slf4j.Slf4j;
@@ -146,6 +145,9 @@ public class CollectionLogHelperPlugin extends Plugin
 	private TempleSyncOrchestrator templeSyncOrchestrator;
 
 	@Inject
+	private CollectionLogNetImportOrchestrator collectionLogNetImportOrchestrator;
+
+	@Inject
 	private ChatEventHandler chatEventHandler;
 
 	@Inject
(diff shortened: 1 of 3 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#533
**Files changed**: 3 (+318/-36)
**Reviewer focus**: behavior parity with the original inline callback, executor lifecycle, thread discipline, and the 5 claimed test paths.

### Findings

#### CRITICAL (0)
None.

#### HIGH (0)
None.

#### MEDIUM (1)
- `requestImport_successfulResultTriggersRebuild` only asserts the toast message is non-empty rather than the exact string produced by `ImportResult.success(42).toToastMessage()`. Consider tightening to `assertThat(msg.getValue()).contains(\"42\")` or matching the exact toast so a regression in `toToastMessage()` for the success path would be caught here. Other tests already pin the exact strings for the non-success branches.

#### LOW (2)
- Javadoc on `requestImport` says `panel` "Must be non-null", while `TempleSyncOrchestrator.requestSync` documents the panel as nullable during shutdown races. Behavior parity with the original (which didn't null-check) is preserved, but it's worth aligning the two javadocs so future readers don't have to compare them — either both nullable, or both non-null with a guarded call site.
- `catch (Exception e)` in `awaitResult` is broad. This matches the original inline behavior verbatim (intentional fail-soft), so it is not a change here, just a note that the next refactor pass on this seam could narrow it to `InterruptedException | ExecutionException`.

### Verification ran
- [x] `./gradlew test --tests CollectionLogNetImportOrchestratorTest` — 5/5 passed
- [x] CI: both `build` checks pass on the head commit
- [x] Diff-level behavior parity audit vs original inline callback (player-name resolution, empty guard, importer call, executor submit, toast, rebuild on success, catch behavior) — all preserved
- [x] Executor lifecycle: `panel` + `httpResultExecutor` are passed per-call, mirroring `TempleSyncOrchestrator.requestSync` and `LootSyncManager` — confirmed
- [x] Thread discipline: async work submitted to the daemon executor; no EDT/client-thread blocking introduced — confirmed
- [x] All 5 test paths present and meaningful (no local player, empty name, success+rebuild, non-success no-rebuild, future failure)

### Recommendation
**SHIP** — no CRITICAL or HIGH issues. The MEDIUM is a test-strength suggestion that does not need to block merge, and the LOWs are doc/style nits.

### Notes
- Plugin LOC delta (-27) and new orchestrator (~130 LOC including license + javadoc) are consistent with the Wave 9 `TempleSyncOrchestrator` extraction shape.
- Worth noting for the #503 tracker: the inline-style comment in the spec ("refreshFromCollectionState + activateGuidance") does not match what the original code actually did (just `panel.rebuild()`). Delta-2 correctly preserved the real behavior — that should be reflected in the spec for the next wave so reviewers don't get whiplash.
- The `doAnswer` inline-executor test trick is a clean pattern; consider hoisting it into a small test helper if Wave 12 ends up using the same shape.

PR #535, opened 2026-05-18
refactor: extract StepChangeHandler from GuidanceOverlayCoordinator (#503)

cha-ndler:
## Summary
- Extracts the InfoBox progress refresh + panel step-progress push from `GuidanceOverlayCoordinator.onStepChanged` into a new `StepChangeHandler` Guice singleton (sibling under `com.collectionloghelper.guidance`).
- Pure extraction: no behavioural change. The overlay clear + apply + NPC-tracker scan stays in the coordinator because those paths mutate coordinator-owned state (`activeMapPoint`, `lastMessagedStepIndex`).
- `activeInfoBox` is mutated in-place via `setStepText` / `setTooltipText` / `setTextColor`, so the handler needs no result-write-back -- only an `Input` carrying `(activeInfoBox, panel, activeTargetItemId)`.

## LOC impact
- `GuidanceOverlayCoordinator.java`: **667 -> 627** (-40). New file `StepChangeHandler.java` at 168 LOC.
- Wave 11 of #503. Falls short of the 600 target by 27; the extracted seam was ~47 LOC honest, not the ~60 originally predicted.

## Performance
- `onStepChanged` fires on step transitions (not per-tick) but can fire repeatedly during fast-skip / cascade. Both guard-fail branches (`activeInfoBox == null`, `panel == null`) allocate nothing -- the only allocations are the `Input` record per call and the existing `Collections.emptyList()` literals.

## Test plan
- [x] `./gradlew build` -- BUILD SUCCESSFUL
- [x] `./gradlew test` -- all green, including the four reflectively-instantiated coordinator tests (`GuidanceOverlayCoordinatorMapPointTest`, `GuidanceOverlayCoordinatorNpcIdTest`, `GuidanceOverlayCoordinatorDescriptionTest`, `DynamicTargetEvaluatorDispatchTest`)
- [x] Coordinator constructor signature extended with `StepChangeHandler`; tests inject a constructor-built singleton via the existing `newXxx` helper pattern

### DIFF
--- a/src/main/java/com/collectionloghelper/guidance/GuidanceOverlayCoordinator.java
+++ b/src/main/java/com/collectionloghelper/guidance/GuidanceOverlayCoordinator.java
@@ -117,6 +117,7 @@ public class GuidanceOverlayCoordinator
 	private final WorldMapController worldMapController;
 	private final DynamicTargetManager dynamicTargetManager;
 	private final NpcTrackerHelper npcTrackerHelper;
+	private final StepChangeHandler stepChangeHandler;
 
 	// -- Guidance UI state (previously in plugin) --
 
@@ -188,7 +189,8 @@ public class GuidanceOverlayCoordinator
 		OverlayDeactivator overlayDeactivator,
 		WorldMapController worldMapController,
 		DynamicTargetManager dynamicTargetManager,
-		NpcTrackerHelper npcTrackerHelper)
+		NpcTrackerHelper npcTrackerHelper,
+		StepChangeHandler stepChangeHandler)
 	{
 		this.client = client;
 		this.clientThread = clientThread;
@@ -217,6 +219,7 @@ public class GuidanceOverlayCoordinator
 		this.worldMapController = worldMapController;
 		this.dynamicTargetManager = dynamicTargetManager;
 		this.npcTrackerHelper = npcTrackerHelper;
+		this.stepChangeHandler = stepChangeHandler;
 	}
 
 	/**
@@ -573,6 +576,10 @@ public class GuidanceOverlayCoordinator
 
 	/**
 	 * Callback from GuidanceSequencer when the current step changes.
+	 * Performs the overlay clear + apply + NPC-tracker scan locally (those
+	 * paths mutate coordinator-owned state such as {@code lastMessagedStepIndex}
+	 * and {@code activeMapPoint}), then delegates the InfoBox/panel
+	 * progress refresh to {@link StepChangeHandler}.
 	 */
 	private void onStepChanged(GuidanceStep step)
 	{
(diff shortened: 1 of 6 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#535
**Files changed**: 6 (+226/-58)
**Reviewer focus**: behavior parity of the `onStepChanged` extraction, in-place mutation semantics of `activeInfoBox`, allocation discipline on hot guard-fail paths, coordinator state-ownership boundary, reflective test harness consistency.

### Findings

#### CRITICAL (0)
- None.

#### HIGH (0)
- None.

#### MEDIUM (2)

1. **Closure capture of `panel` differs subtly from pre-extraction.** In master the `clientThread.invokeLater` lambda inside `onStepChanged` referenced the coordinator field `panel` directly, so it read the field value at lambda-execution time. The extracted handler snapshots `in.panel` into `final CollectionLogHelperPanel panelRef` at call time and the lambda captures that snapshot. In practice the `panel` field is set once at plugin startup via `@Setter setPanel(...)` and is never replaced or nulled while a guidance sequence is active, so this is benign. Worth flagging because the PR is presented as a pure extraction with no behavioural change. Consider a one-line comment in `StepChangeHandler.handle` explaining the snapshot is intentional and relying on the panel-set-once invariant — otherwise a future maintainer who adds panel-rebuild logic could trip on a stale reference.

2. **`Input` is a natural candidate for a Java `record`.** The handler is on JDK 17, the `Input` class has three final nullable fields and a single constructor, no behaviour. A `record Input(@Nullable GuidanceInfoBox activeInfoBox, @Nullable CollectionLogHelperPanel panel, @Nullable Integer activeTargetItemId) {}` would save ~12 LOC and match the project's stated Java style preference for records on value types (`.claude/rules/java/coding-style.md`). Sibling `*.Input` types in this package (e.g. `OverlayStepApplier.Input`, `OverlayDeactivator.Input`) appear to use classes too, so this is consistency-with-siblings vs. consistency-with-style-guide — call it as you prefer, but a follow-up sweep to records across the guidance package would be reasonable.

#### LOW (3)

1. **`import java.awt.Color`** is now used directly in `StepChangeHandler`, replacing the fully-qualified `java.awt.Color.GREEN` reference that was in the coordinator. Good — cleaner import discipline. No action.

2. **Javadoc on `handle` documents in-place mutation explicitly** ("mutates in.activeInfoBox in-place; no field reassignment needed back on the coordinator"). This is the right place to document the contract since the in-place mutation is non-obvious for a reviewer skimming the call site. Nice.

3. **Coordinator-target LOC**: 627 vs. 600 stated target. The PR body is upfront about this (27 over). After spot-checking the file, `applyDynamicItemObjectOverlays` (lines 493~ to ~550, est. ~55 LOC) is indeed the natural next seam — it builds `matchedObjectIds` / `matchedItemIds` / `tooltipText` / `action` purely from the current step's `dynamicItemObjectTiers` and `playerInventoryState`, then pushes to overlays. The overlay push is coordinator-touching but the tier-resolution half is a clean pure function that could move to a `DynamicItemObjectTierResolver` and bring the coordinator under 600. No need to do it in this PR.

### Verification ran
- [x] `./gradlew test` — 1222 tests, 0 failures, 0 errors
- [x] `./gradlew build` — BUILD SUCCESSFUL
- [x] Reflective test harnesses (`MapPointTest`, `NpcIdTest`, `DescriptionTest`, `DynamicTargetEvaluatorDispatchTest`) all extend the constructor signature and `newStepChangeHandler()` helper consistently across the four sibling tests
- [x] State-ownership boundary holds: `applyStepToOverlays` writes back `lastMessagedStepIndex` + `activeMapPoint` via `OverlayStepApplier.Result`, and `npcTrackerHelper.scanForTrackedNpc` reads coordinator-owned tracker state — both correctly stayed in the coordinator
- [x] Hot-path allocation check: both guard-fail branches in `StepChangeHandler.handle` (activeInfoBox-null path and panel-null path) allocate nothing; the only per-call allocations are the `Input` instance at the call site and the existing `Collections.emptyList()` literals carried over from master
- [x] In-place mutation preserved: `setStepText` / `setTooltipText` / `setTextColor` all mutate the existing `GuidanceInfoBox` instance; no `activeInfoBox = new ...` write-back, so the coordinator field stays current without a Result type
- [x] Behaviour parity: order of operations within `onStepChanged` is `clearStepOverlays -> applyStepToOverlays -> scanForTrackedNpc -> handle(InfoBox + panel push)`, matching master line-for-line

### Recommendation
**SHIP** — pure extraction, no CRITICAL/HIGH issues. The 2 MEDIUMs are stylistic/defensive — neither blocks merge. The handler is well-scoped, the in-place mutation contract is documented at the right place, and the test harness changes are mechanical and consistent across the four reflective tests. Wave 11 is in good shape.

### Notes

- Coordinator is *close* to architectural floor at 627 LOC but not at it. The honest read after spot-checking the file is that `applyDynamicItemObjectOverlays` is the next viable seam (~55 LOC of tier resolution that can move to a `DynamicItemObjectTierResolver`), which would land the coordinator at ~570. After that the remaining LOC is mostly field declarations, the long constructor parameter list (32 deps), and the small dispatch methods — those are not seams, they are the irreducible facade. The 600-LOC target is reachable in one more wave.

- Performance follow-up (#523/#527): worth confirming during cascade testing that the per-call `Input` allocation does not register on the `onStepChanged` profile. With step transitions firing at most a handful of times per second even under fast-skip, a single 3-field allocation is in the noise, but a sibling-style "lean Input via static method args" alternative exists if a future profile flags it.

- Checklist gap worth recording for future agent runs: the project's Java style guide prefers `record` for value-type bags, but the guidance package consistently uses inner-class `Input` instead. Either choose one and apply uniformly in a follow-up sweep, or document the deviation in the package README.

PR #530, opened 2026-05-18
refactor: extract PlayerLocationResolver from Plugin (#503)

cha-ndler:
## Summary

Wave 10 of the #503 god-class refactor. Extracts `resolvePlayerWorldLocation` and its companion `cachedPlayerLocation` volatile field out of `CollectionLogHelperPlugin` into a new Guice-managed `lifecycle/PlayerLocationResolver`.

The resolver owns the cache and exposes:
- `resolveAndCache()` — computes the player's real-world `WorldPoint` (handling instanced regions + non-top-level WorldViews like sailing boats), updates the cache, returns the value.
- `getCachedLocation()` — `@Nullable` volatile read for cross-thread callers.
- `reset()` — clears the cache on logout / profile change.

## LOC delta

`CollectionLogHelperPlugin.java`: **866 -> 780 (-86 LOC)**. Closer to the 600-LOC target for #503.

## Behavior

No behavior change. The resolver method body is the existing logic moved verbatim. The four removed `cachedPlayerLocation` reads/writes in Plugin now go through the resolver's `resolveAndCache()`, `getCachedLocation()`, and `reset()` (twice — on profile change and on login-screen state).

## Test plan

- [x] `./gradlew build` passes
- [x] `./gradlew test` passes (1210 tests)
- [x] `OnSequenceCompletePerItemTest` updated to inject the new `playerLocationResolver` mock so the captured runnable does not NPE
- [ ] Manual smoke: ARRIVE_AT_TILE auto-advance still works in an instanced region (e.g. Royal Titans)
- [ ] Manual smoke: location resolves correctly while sailing on a `WorldEntity`

PR #536, opened 2026-05-18
test: fix PanelRebuildOrchestratorTest scrollbar flake on JDK 11 and JDK 17

cha-ndler:
## Summary

The `PanelRebuildOrchestratorTest` scrollbar tests have flaked intermittently on headless Linux CI across **both JDK 11 and JDK 17**. Previously diagnosed as JDK-17-only (#515, #516) and patched with a `pinScrollBounds` helper that called `bar.setValues(0, 400, 0, 5000)`, the flake recurred today on the master CI run for the #530/#531 merge commits — this time on JDK 11.

### Root cause

The earlier `pinScrollBounds(bar)` wrote value+extent+min+max onto the model, but **a subsequent Swing layout pass on the viewport would re-narrow the BoundedRangeModel** (back to viewport-derived `max=100, extent=10`) between the test's `setValue(...)` and the assertion. That re-narrowing clamped values like `250` down to `90`. The flake is racy because it depends on when the EDT layout pass runs relative to the test thread's reads — which is JDK-version and headless-Linux-load sensitive.

### Fix (Option B — pinned-bounds model)

Started with Option A (atomic `setRangeProperties`), but A doesn't actually address the race: a *later* `setMaximum(100)` from a layout pass still clamps. Escalated to Option B as the brief allowed.

Replace the vertical scrollbar's model with a `PinnedRangeModel`: a `DefaultBoundedRangeModel` subclass that, once `pin()`-ed, silently rejects any `setMinimum`/`setMaximum`/`setExtent` or range-narrowing `setRangeProperties` call while still honouring `setValue(...)` writes. `pinScrollBounds(...)` now swaps in this model on first use and writes value+range atomically via `setRangeProperties`; further re-pins are idempotent.

This is a **test-only change** — `PanelRebuildOrchestrator` production code is untouched.

## Test plan

- [x] Local clean rebuild + 15 consecutive test-class reruns on Windows JDK 17 — all green
- [x] Full local `./gradlew test` — all green
- [ ] CI Build (Linux JDK 11) green
- [ ] CI CI (Linux JDK 17) green

### DIFF
--- a/src/test/java/com/collectionloghelper/ui/PanelRebuildOrchestratorTest.java
+++ b/src/test/java/com/collectionloghelper/ui/PanelRebuildOrchestratorTest.java
@@ -33,6 +33,7 @@ import java.awt.Dimension;
 import java.util.Collections;
 import java.util.Set;
 import javax.swing.BoxLayout;
+import javax.swing.DefaultBoundedRangeModel;
 import javax.swing.JLabel;
 import javax.swing.JPanel;
 import javax.swing.JScrollBar;
@@ -91,14 +92,22 @@ public class PanelRebuildOrchestratorTest
 		hostPanel.add(listContainer, BorderLayout.NORTH);
 
 		scrollPane = new JScrollPane(hostPanel);
+		// Swap the vertical scrollbar's model for one that ignores any
+		// subsequent narrowing of its bounds (extent/min/max). This is the
+		// bulletproof fix for the headless-Linux flake where viewport layout
+		// passes re-clamp value=250 down to value=90 between setValue(...) and
+		// the test's assertion. The pinned model accepts setValue freely but
+		// refuses to shrink its own range after pinScrollBounds(...) has been
+		// called. See also the comment on pinScrollBounds below.
+		scrollPane.setVerticalScrollBar(new JScrollBar(JScrollBar.VERTICAL));
 		scrollPane.setSize(200, 400);
 		scrollPane.doLayout();
 		scrollPane.getViewport().doLayout();
 		// Pin the vertical scrollbar's model bounds explicitly so setValue(...) is
 		// not clamped by lazy/late viewport layout. Headless CI environments
-		// (Linux, JDK 17) have been observed to leave the model at its default
-		// (max=100, extent=10) when assertions read the value immediately after
-		// setValue, causing values like 123 to clamp to 90.
+		// (Linux, JDK 11 and JDK 17) have been observed to leave the model at its
+		// default (max=100, extent=10) when assertions read the value immediately
+		// after setValue, causing values like 250 to clamp to 90.
 		pinScrollBounds(scrollPane.getVerticalScrollBar());
 
 		orchestrator = new PanelRebuildOrchestrator(hostPanel, listContainer);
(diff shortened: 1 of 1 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#536
**Files changed**: 1 (+108/-11)
**Reviewer focus**: Scope discipline, `PinnedRangeModel` correctness (pin-gating + value pass-through), idempotency of `pinScrollBounds`, behavioural fidelity of pinned tests, and dead-helper cleanup.

### Findings

#### CRITICAL (0)
- None.

#### HIGH (0)
- None.

#### MEDIUM (2)
- **Pinned-model tests trend toward tautology.** Once the model is pinned to `[0, 5000]` with extent `400`, every `setValue(n)` for `n in [0, 4600]` round-trips unchanged. Tests like `captureRecordsScrollPositionFromEnclosingScrollPane` (asserts `250 == snapshot.getScrollPosition()`) and `deferScrollRestoreSchedulesValueOnVerticalScrollBar` (asserts `400 == vBar().getValue()` after defer) no longer prove anything about scrollbar clamping mechanics; they prove the orchestrator reads from / writes to the JScrollBar the production code is wired to. That is still a useful integration check, but the test names and comments imply something stronger. Consider a brief comment in the test class header explicitly noting that scroll-value tests are *plumbing checks*, not *clamping checks*, so a future reader does not mistake green tests for evidence that real scrollbar clamp behaviour works.
- **Constants for the pinned range.** `0, 400, 0, 5000` is repeated in `pinScrollBounds` and matches assumptions buried inside the clamp arithmetic. Naming them (`PINNED_INITIAL_VALUE = 0`, `PINNED_EXTENT = 400`, `PINNED_MAX = 5000`) would keep the helper self-documenting and avoid the magic-number smell flagged by `.claude/rules/common/coding-style.md`.

#### LOW (3)
- **`pinScrollBounds` idempotency leans on the override.** Re-calling the helper invokes `model.setRangeProperties(0, 400, 0, 5000, false)` *after* `pinned=true`, which goes through the overridden path and silently discards the supplied range params, keeping the existing `[0,5000]/400`. This is correct but subtle; a one-liner in the helper noting "re-pin is a no-op aside from resetting value to 0" would aid the next reader.
- **Initial `JScrollBar` swap (line 102) is somewhat redundant.** `JScrollPane` already builds a vertical scrollbar; replacing it with a fresh `JScrollBar(JScrollBar.VERTICAL)` only to immediately replace its model in `pinScrollBounds` is one extra step in the setup. Functionally fine, but slightly noisier than calling `pinScrollBounds(scrollPane.getVerticalScrollBar())` directly on the pane-supplied bar. Not blocking.
- **Clamp expression is conservative.** `Math.max(getMinimum(), Math.min(newValue, getMaximum() - getExtent()))` mirrors `DefaultBoundedRangeModel`'s own clamp; it would be slightly tidier to delegate to `super.setRangeProperties(newValue, getExtent(), getMinimum(), getMaximum(), adjusting)` and let the superclass clamp, since it already does. Functionally equivalent.

### Verification ran
- [x] `./gradlew test --tests PanelRebuildOrchestratorTest` (JDK 17) — green
- [x] `./gradlew test --tests PanelRebuildOrchestratorTest --rerun-tasks` (JDK 17) — green
- [x] `git diff master..HEAD --stat` — single test file modified, zero production-code touch
- [x] CI rollup confirms Build (JDK 11) + CI (JDK 17) both green on this branch
- [x] Old `bar.setValues(0, 400, 0, 5000)` body confirmed replaced (not left as dead code)
- [x] `PinnedRangeModel` correctness walk-through:
  - default-uninitialised state allows the setup `setRangeProperties(0, 400, 0, 5000, false)` before `pin()`
  - after `pin()`, `setMinimum/setMaximum/setExtent` all no-op
  - `setRangeProperties` after pin keeps existing range and clamps `newValue` only — verified `setValue(250)` and `setValue(400)` paths
  - re-calling `pinScrollBounds` is idempotent

### Recommendation
**SHIP** — no CRITICAL/HIGH issues. The fix is genuinely robust against the documented race (subsequent layout-pass `setMaximum(100)` cannot re-narrow), correctly diagnoses why the earlier `bar.setValues(...)` patch was insufficient, and is scoped strictly to test code. The MEDIUM items are clarity-of-intent suggestions, not blockers, and can be addressed in a follow-up or ignored.

### Notes
- The cross-JDK flake recurrence (originally diagnosed JDK-17-only in #515/#516, now observed on JDK 11 too) is a useful reminder: when a fix is keyed to a JDK version in a comment or commit, the next reader will assume that version-bound scope is real. Future test-flake fixes should prefer environment-agnostic language unless we have evidence that the underlying mechanism is genuinely JDK-version-specific. Worth a one-line addition to `feedback_flaky_test_patterns.md`.
- The Option-A vs Option-B reasoning in the PR body is exemplary: the agent identified that atomic `setRangeProperties` only fixes intermediate-state observation, not a *subsequent* `setMaximum(100)`. That distinction is exactly the kind of thing that gets missed when "atomic write" feels like a sufficient answer. Recommend preserving that reasoning in a memory note for the next overlay/test-flake encounter.

PR #538, opened 2026-05-18
refactor: extract VarbitChangeRouter from Plugin (#503)

cha-ndler:
## Summary

Wave 12 of the #503 god-class split. Extracts the body of `CollectionLogHelperPlugin.onVarbitChanged` into a focused `VarbitChangeRouter` collaborator under `com.collectionloghelper.lifecycle`, mirroring the `ChatEventHandler` pattern from #525.

The plugin keeps the `@Subscribe` annotation (so the RuneLite event bus still discovers the handler via `eventBus.register(plugin)`); the body becomes a single `varbitChangeRouter.handle(event)` delegation.

Three narrow `Runnable` callbacks keep the per-tick refresh flags and rebuild triggers as plugin-owned state without leaking those fields into the router:
- `flagRefreshes` — sets `pendingRequirementsRefresh` + `pendingTravelVarbitRefresh`
- `onCollectionStateChanged` — calls `rebuildSourcesWithMissingItems()`
- `onSlayerTaskChanged` — sets `pendingPanelRebuild` + `rankedSourcesDirty`

## LOC delta

- `CollectionLogHelperPlugin.java`: **753 -> 719** (-34 net)
- New `VarbitChangeRouter.java`: 187 LOC (license + javadoc + class)
- New `VarbitChangeRouterTest.java`: 257 LOC

## Performance note

`VarbitChanged` fires extremely frequently in OSRS (hundreds of times during login alone, see #527 / #523 lesson). The no-op / guard-fail path stays allocation-free — no per-event `new` sites, no captured lambdas, no boxing. The `setCallbacks` Runnables are wired once in `startUp()`.

## Test plan

- [x] New `VarbitChangeRouterTest` covers:
  - Slayer-task delta detection (active->inactive, creature change, remaining change, no-change)
  - Authoring-log throttle (tick advance fires, same tick suppresses, disabled config suppresses)
  - Script-scan guard short-circuits before obtained-count check
  - Obtained-count delta fires collection-state + slayer callbacks
  - Null-callback safety (no NPE if `setCallbacks` not yet wired)
  - Sequencer forwarding always happens, even mid-scan
- [ ] In-game smoke: log in, complete a Slayer task, verify panel rebuild fires
- [ ] In-game smoke: enable `guidanceAuthoring`, verify one `VARBIT_CHANGED` log line per tick

### DIFF
--- a/src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java
+++ b/src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java
@@ -41,6 +41,7 @@ import com.collectionloghelper.lifecycle.OverlayRegistry;
 import com.collectionloghelper.lifecycle.PlayerLocationResolver;
 import com.collectionloghelper.lifecycle.SceneEventRouter;
 import com.collectionloghelper.lifecycle.SyncStateCoordinator;
+import com.collectionloghelper.lifecycle.VarbitChangeRouter;
 import com.collectionloghelper.ui.CollectionLogHelperPanel;
 import com.google.inject.Provides;
 import java.awt.image.BufferedImage;
@@ -152,6 +153,9 @@ public class CollectionLogHelperPlugin extends Plugin
 
 	@Inject
 	private PlayerLocationResolver playerLocationResolver;
+
+	@Inject
+	private VarbitChangeRouter varbitChangeRouter;
 
 
 	private CollectionLogHelperPanel panel;
@@ -294,6 +298,18 @@ public class CollectionLogHelperPlugin extends Plugin
 				pendingPanelRebuild = true;
 				rankedSourcesDirty = true;
 			});
+		varbitChangeRouter.setCallbacks(
+			() ->
+			{
+				pendingRequirementsRefresh = true;
+				pendingTravelVarbitRefresh = true;
+			},
+			this::rebuildSourcesWithMissingItems,
+			() ->
+			{
+				pendingPanelRebuild = true;
+				rankedSourcesDirty = true;
+			});
 		chatCommandManager.registerCommand("clh", chatEventHandler::onClhCommand);
 		log.info("Collection Log Helper started");
 	}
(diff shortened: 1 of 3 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#538
**Files changed**: 3  (+461/-51)
**Reviewer focus**: hot-path allocation discipline in `onVarbitChanged`, callback wiring correctness, slayer-task delta preservation, test coverage assertions

### Findings

#### CRITICAL (0)
None.

#### HIGH (0)
None.

#### MEDIUM (1)

- **Test count mismatch in PR body.** Body claims 15 tests; JUnit XML reports `tests="13"`. Not a correctness issue, just stale prose. Worth updating before squash so the merge commit message is accurate, or leaving as-is and noting in CHANGELOG.

#### LOW (2)

- **Method-level @Slf4j is unused.** `VarbitChangeRouter` carries `@Slf4j` but the class body never logs through it — all logging is delegated through `authoringLogger`. Either drop the annotation to remove the synthetic `log` field, or use it for a defensive `log.debug` somewhere (not required). Cosmetic.
- **Public class constructor is package-private.** `@Inject VarbitChangeRouter(...)` is package-private while the class is `public`. Consistent with the rest of `lifecycle/` (e.g., `SyncStateCoordinator`), so this is fine — flagging only because Guice could instantiate either way and explicit `public` would be marginally more conventional. No change required.

### Hot-path verification

Walked `handle()` line-by-line against the no-op / guard-fail path (unrelated varbit, no slayer change, script-scan-active short-circuit):

- No `new` allocations in `handle()` itself.
- The three `Runnable` fields (`flagRefreshes`, `onCollectionStateChanged`, `onSlayerTaskChanged`) are wired **once** in `CollectionLogHelperPlugin.startUp()` — confirmed in the diff. Lambdas are captured once at registration, not per-event.
- `event.getVarbitId()` / `event.getValue()` are only read on the authoring-log branch (guarded behind `config.guidanceAuthoring()`), so the format-string vararg boxing on `authoringLogger.log("VARBIT_CHANGED varbitId=%d value=%d", ...)` only happens at most once per tick when authoring is enabled. Acceptable.
- `getCreatureName().equals(oldCreature)` does a string compare on each event, but that's preserved from the original code — not a regression.
- Guidance sequencer forward (`guidance.getGuidanceSequencer().onVarbitChanged(...)`) is unchanged from the original.

Result: allocation-free hot path on the dominant code path matches the PR description.

### Slayer-delta preservation

Compared the extracted block in `VarbitChangeRouter.handle()` (lines ~140-160 of the new file) against the original `CollectionLogHelperPlugin.onVarbitChanged` block (in the `-` diff hunks). Identical:

1. Capture `wasActive` / `oldCreature` / `oldRemaining` BEFORE `refresh()` — order preserved.
2. Composite OR with null-guard `getCreatureName() != null && !equals(oldCreature)` — preserved including the null short-circuit.
3. `slayerChanged = true` is set inside the obtained-count delta block, so a new item still forces a slayer-callback fire. Preserved.
4. Final `if (slayerChanged && onSlayerTaskChanged != null)` fires the callback after the obtained-count branch — order matches original.

Behavior is byte-for-byte equivalent on this branch.

### Callback wiring

`CollectionLogHelperPlugin.startUp()` invokes `varbitChangeRouter.setCallbacks(...)` once with three lambdas matching the three plugin-owned flag groups. The new `setCallbacks` call sits next to the existing `chatEventHandler` setup (same wave-12 pattern), and both registrations happen before `eventBus.register(plugin)` would dispatch the first `VarbitChanged`. Safe.

Null-callback safety is exercised by `nullCallbacks_doNotNpe` — handle() defensively null-checks each Runnable before invocation. Defensive but good (covers test isolation and any future ordering surprise during plugin restart).

### Test coverage spot-check

13 tests, meaningful assertions throughout:
- `unrelatedVarbit_stillFlagsRefreshes_andForwardsToSequencer` — verifies `refreshVarps`, sequencer forward, `slayerTaskState.refresh`, and exactly which callbacks fire/don't fire.
- Slayer-delta block has 4 tests (active->inactive, creature change, remaining change, no-change) — covers each branch of the OR.
- Authoring-log throttle has 3 tests (tick-advance fires, same-tick suppresses, config-disabled suppresses).
- `scriptScanActive_shortCircuitsBeforeObtainedCheck` uses `verify(collectionState, never()).getTotalObtained()` — confirms the early return order.
- `obtainedCountDelta_firesCollectionAndSlayerCallbacks` asserts both callbacks fire when a new item lands (preserves the `slayerChanged = true; // rebuild anyway` semantic).

Stronger than usual for a refactor PR.

### Verification ran
- [x] `./gradlew test --tests VarbitChangeRouterTest` — 13/13 passed
- [x] CI build + CI workflow — both green
- [x] Slayer-delta block byte-compared against pre-extraction inline body — identical

### Recommendation
**SHIP** — no CRITICAL/HIGH issues. The MEDIUM (test-count prose mismatch) and LOWs (unused @Slf4j, ctor visibility) are cosmetic.

### Notes
- Pattern continues to look clean across the wave-12 god-class split — `VarbitChangeRouter` mirrors `ChatEventHandler` (#525) and `SceneEventRouter` very closely. Consider adding the "Runnable callback wiring vs. constructor injection" rationale (avoid Guice cycle on the plugin) to the in-repo refactor notes if not already documented; it's a recurring pattern now and would help future waves.
- The `// rebuild anyway` semantic — where an obtained-count delta forces `slayerChanged = true` — is preserved but not super-discoverable. A one-line comment in `handle()` explaining *why* an item delta forces a panel rebuild (because ranked-source ordering depends on remaining items) would aid the next reader. Not a blocker.
- Suggest the wave-13 PR continue with the remaining inline handlers (`onChatMessage` is already done, `onScriptPostFired`, `onItemContainerChanged`, etc. — whichever is next per roadmap).

PR #539, opened 2026-05-18
perf: avoid per-tick allocations in DynamicTargetManager guard-fail path (#527)

### LINKED ISSUE #527: perf: avoid per-tick allocations in DynamicTargetManager guard-fail path (#523 follow-up)

cha-ndler:
## Summary

Fixes #527. PR #523 extracted `DynamicTargetManager` but introduced a per-tick allocation regression on guard-fail paths. Every game tick the coordinator allocated a fresh `Input` carrier and the manager allocated a fresh `Result` wrapper, even when guidance was inactive (the dominant case).

## Approach — both options from the issue, combined

**Option 1 (primary)** — lifted the `!isActive()` guard back to the coordinator call site. Inactive ticks now skip the manager entirely and allocate nothing.

**Option 2 (secondary)** — collapsed the manager API: `tick()` now takes its inputs as method arguments and returns `@Nullable CollectionLogWorldMapPoint` (`null` == "no change"). The `Input` and `Result` nested classes are deleted, so the three remaining active-tick guard-fails (`step == null` / unregistered evaluator / evaluator returns `null`) also allocate nothing.

Option 1 alone only handled the inactive-tick case; the active-tick guard-fails still happen frequently (any step that doesn't carry a `dynamicTargetEvaluator` key), so Option 2 covers the rest.

## Allocation profile

| Tick path | Before | After |
|---|---|---|
| Inactive guidance | 1 × `Input`, 1 × `Result` | **0** |
| Active, step has no evaluator | 1 × `Input`, 1 × `Result` | **0** |
| Active, evaluator key unregistered | 1 × `Input`, 1 × `Result` | **0** |
| Active, evaluator returns `null` | 1 × `Input`, 1 × `Result` | **0** |
| Happy path (new target produced) | 1 × `Input`, 1 × `Result`, 1 × `CollectionLogWorldMapPoint` | 1 × `CollectionLogWorldMapPoint` |

## Test plan

- [x] New `DynamicTargetManagerTest` covers all four guard-fail paths (asserts `null` return + zero overlay interactions) plus two happy-path scenarios
- [x] Existing `DynamicTargetEvaluatorDispatchTest` unchanged and still green — continues to cover coordinator-side dispatch including the inactive-guidance no-op
- [ ] Manual smoke: start guidance, advance through a few steps with and without `dynamicTargetEvaluator` keys, confirm world map / minimap / hint arrow behaviour unchanged

Closes #527.

### DIFF
--- a/src/main/java/com/collectionloghelper/guidance/DynamicTargetManager.java
+++ b/src/main/java/com/collectionloghelper/guidance/DynamicTargetManager.java
@@ -45,11 +45,10 @@ import net.runelite.client.ui.overlay.worldmap.WorldMapPointManager;
  *
  * <p>Pure extraction from {@link GuidanceOverlayCoordinator}: this manager
  * holds no mutable state of its own.  Coordinator state it needs to read
- * (active map point, current guidance step, sequencer activity, active
- * target item id, cached log icon) is passed in via the {@link Input}
- * carrier, and any state mutation the coordinator must apply (the new
- * active map point) is returned in {@link Result} so the coordinator
- * remains the single owner of guidance state.</p>
+ * (active map point, current guidance step, active target item id, cached
+ * log icon) is passed in as method arguments, and any state mutation the
+ * coordinator must apply (the new active map point) is returned directly
+ * so the coordinator remains the single owner of guidance state.</p>
  *
  * <p>Responsibilities:
  * <ul>
@@ -60,16 +59,22 @@ import net.runelite.client.ui.overlay.worldmap.WorldMapPointManager;
  *   <li>Pushing the new target into the four location-sensitive overlays
  *       (in-world, minimap, world-map route, world-map destination).</li>
  *   <li>Refreshing the world map point in {@link WorldMapPointManager} and
- *       reporting the new map point back via {@link Result}.</li>
+ *       returning the new map point to the coordinator.</li>
  *   <li>Re-applying the hint arrow via {@link WorldMapController} when
  *       {@code config.showHintArrow()} is enabled and the local player is
  *       on a compatible plane / world view.</li>
  * </ul>
  *
- * <p>Called on the client thread once per game tick.  Returns silently
- * (with {@link Result#unchanged(CollectionLogWorldMapPoint)}) when guidance
- * is inactive, the step has no evaluator key, the key is unregistered, or
- * the evaluator returns {@code null}.</p>
+ * <p>Called on the client thread once per game tick <em>only when guidance
+ * is active</em> (the coordinator guards the call site).  Returns
+ * {@code null} when the step has no evaluator key, the key is unregistered,
+ * or the evaluator returns {@code null}; the coordinator treats {@code null}
+ * as "no change" and keeps its existing active map point.</p>
+ *
+ * <p>This signature is allocation-free on guard-fail paths (issue #527):
+ * inactive ticks bypass the call entirely (Option 1 — coordinator-side
+ * guard), and the three active-tick guard-fails return a cached {@code null}
+ * sentinel without constructing any wrapper objects.</p>
  */
 @Slf4j
 @Singleton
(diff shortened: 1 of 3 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#539
**Files changed**: 3  (+334/-108)
**Reviewer focus**: per-tick allocation profile of `DynamicTargetManager.tick()` on all four guard-fail paths and the happy path; behavior parity with pre-#523 master; coordinator-side `isActive()` guard correctness; test coverage of the new direct-args signature.

### Findings

#### CRITICAL (0)
- None.

#### HIGH (0)
- None.

#### MEDIUM (0)
- None.

#### LOW (2)
- `DynamicTargetManagerTest.tick_evaluatorReturnsPoint_pushesToOverlaysAndReturnsNewMapPoint` does not assert `verify(worldMapDestinationOverlay).setTarget(...)`. The happy path covers `setMapPointActive(true)` but the preceding `setTarget` call is not directly verified. Not strictly necessary because the `setMapPointActive(true)` verification gates the same code branch, but adding `verify(worldMapDestinationOverlay).setTarget(eq(DYNAMIC_POINT), any());` would tighten the contract.
- `verifyNoOverlayInteractions()` uses `verify(worldMapDestinationOverlay, never()).setMapPointActive(true);` — narrow predicate (`true` only). If a future regression flipped to `setMapPointActive(false)` on a guard-fail path it would slip through. Consider `setMapPointActive(anyBoolean())` for stronger no-interaction guarantees.

### Allocation-profile verification

Walked `DynamicTargetManager.tick()` end-to-end against the claim:

| Tick path | Pre-fix | Post-fix | Verified |
|---|---|---|---|
| Inactive guidance | 1 × `Input`, 1 × `Result` | 0 (guard at call site, manager not called) | yes — `GuidanceOverlayCoordinator.tick()` line 409 guard |
| Active, `currentStep == null` | 1 × `Input`, 1 × `Result` | 0 (early `return null`) | yes — line 137-139 |
| Active, `dynamicTargetEvaluator == null` | 1 × `Input`, 1 × `Result` | 0 (early `return null`) | yes — line 137-139 |
| Active, evaluator unregistered | 1 × `Input`, 1 × `Result` | 0 (early `return null`) | yes — line 141-144 |
| Active, evaluator returns `null` | 1 × `Input`, 1 × `Result` | 0 (early `return null`) | yes — line 146-150 |
| Happy path | 1 × `Input`, 1 × `Result`, 1 × `CollectionLogWorldMapPoint` | 1 × `CollectionLogWorldMapPoint` | yes — line 162 |

All four guard-fail paths return the cached `null` literal; no autoboxing, no wrapper. The coordinator dereferences `null` with a `!= null` check before assigning to `activeMapPoint`, so the prior "echo existing point" semantic is preserved by simply *not* reassigning. Behavior parity is intact.

### Coordinator-side guard correctness

`guidanceSequencer.isActive()` reads a non-volatile boolean field. Both the read site (`GuidanceOverlayCoordinator.tick()`) and the writers (`GuidanceSequencer.start()` line 136, `stop()` lines 158/669, `StepAdvancer` lines 188/248) execute on the RuneLite client thread. There is no cross-thread race: a guard-then-call sequence sees a stable `active` value within a single tick. No memory-visibility concern introduced by lifting the guard.

### API simplification correctness

`Input` and `Result` were package-private static nested classes referenced only by `DynamicTargetManager` itself and `GuidanceOverlayCoordinator`. Repo-wide grep for `DynamicTargetManager.(Input|Result)` returns zero hits after the change — no external callers were depending on them. Removal is safe.

### Test coverage

`DynamicTargetManagerTest` (6 tests) covers:
- Guard-fail #1: null step (covered)
- Guard-fail #2: null evaluator key (covered)
- Guard-fail #3: unregistered key (covered)
- Guard-fail #4: evaluator returns null (covered)
- Happy path: evaluator returns point, all overlays pushed (covered)
- Happy path with previous active map point: previous removed (covered)

The inactive-guidance path (Option 1 guard) is not retested here; it lives in `DynamicTargetEvaluatorDispatchTest.tick_inactiveGuidance_noOpForOverlays` (line 278-285 in that file), still present and green. Coverage matrix is complete across both test classes.

### Verification ran
- [x] Targeted: `DynamicTargetManagerTest` + `DynamicTargetEvaluatorDispatchTest` — green
- [x] CI `build` + `CI` workflow checks — both SUCCESS
- [x] Repo-wide grep for stale `DynamicTargetManager.Input` / `.Result` references — zero hits
- [x] Sequencer threading model for `isActive()` guard — client-thread-only mutations, no race

### Recommendation
**SHIP** — no CRITICAL/HIGH/MEDIUM issues. Two LOW polish suggestions on test assertion tightness are optional and not blocking.

### Notes
- Clean closure of #527. The combined Option 1+2 approach is the right call: Option 1 alone would have left 3 of 5 paths still allocating; Option 2 alone would have wasted work on the inactive-tick majority case.
- Javadoc updates (precondition, allocation-discipline rationale) are well-written and will help the next maintainer understand why the call-site guard exists.
- Happy-path output is bit-for-bit identical to pre-fix: same overlays touched in the same order, same `CollectionLogWorldMapPoint` constructor args, same hint-arrow gating.
- Worth recording in roadmap status log that the #523-introduced regression is now fully retired.

PR #540, opened 2026-05-18
refactor: extract CollectionStateChangeHandler from Plugin (#503)

cha-ndler:
## Summary
- Wave 13 of the #503 god-class split. Extracts the `onSequenceComplete + rebuildSourcesWithMissingItems + exportEfficiencyIfEnabled` trio from `CollectionLogHelperPlugin` into a focused `lifecycle.CollectionStateChangeHandler` collaborator.
- `Plugin.java`: **719 -> 678 LOC** (-41). Plugin retains state ownership of `sourcesWithMissingItems`, `pendingPanelRebuild`, `rankedSourcesDirty`, and `cachedRankedSources`; the handler reads/writes those fields exclusively via three narrow callbacks wired once in `startUp()`.
- `GuidanceOverlayCoordinator.setOnSequenceCompleteCallback` now points at `collectionStateChangeHandler::handleSequenceComplete`, mirroring the Wave 11/12 (`SyncStateCoordinator`, `VarbitChangeRouter`) injection pattern.

## Refactor details
- New `@Singleton` Guice component injected via constructor (`Client`, `CollectionLogHelperConfig`, `DataModule`, `EfficiencyModule`, `GuidanceModule`).
- `setCallbacks(Runnable flags, Supplier<List<ScoredItem>> ranked, BiConsumer<CollectionLogSource, Integer> activate)` keeps the circular plugin dependency off the constructor.
- `handleSequenceComplete` short-circuits with no allocation when auto-advance is disabled (matches the #527 / #523 hot-path discipline used in `VarbitChangeRouter`).
- `rebuildSourcesWithMissingItems` returns the new `Set<String>` so the plugin's `shutDown` reset semantics for `sourcesWithMissingItems` are preserved.

## Test plan
- [x] `./gradlew build` passes (incl. JaCoCo coverage verification + drop-rate lint).
- [x] `./gradlew test` — 1267 tests, all green.
- [x] `OnSequenceCompletePerItemTest` rewritten to drive the handler directly (was using reflection against the deleted private `Plugin.onSequenceComplete`).
- [x] New `CollectionStateChangeHandlerTest` covers: flag-callback always fires, auto-advance disabled path is a no-op, empty/null ranked-sources path, locked-source skip, null-callback safety, coordinator partition delegation, export disabled no-op, and the player-name-unavailable fallback.

## Wave 14 candidate
Eta-2 flagged `onGameTick` (~85 LOC) as the next-biggest seam. Confirmed: lines 497-577 of the post-Wave-13 `Plugin.java` are an ~80 LOC tick-coalescer with five distinct responsibilities (debounced requirements / travel-varbit refresh, per-character data init, guidance coordinator tick, player-location resolve + sequencer dispatch, deferred slayer refresh, sync-tick delegation, and final panel rebuild). A `GameTickOrchestrator` extraction looks tractable and should land Plugin near the 600 LOC target.

### DIFF
--- a/src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java
+++ b/src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java
@@ -37,6 +37,7 @@ import com.collectionloghelper.sync.LootSyncManager;
 import com.collectionloghelper.sync.TempleSyncOrchestrator;
 import com.collectionloghelper.efficiency.ScoredItem;
 import com.collectionloghelper.lifecycle.AuthoringLogger;
+import com.collectionloghelper.lifecycle.CollectionStateChangeHandler;
 import com.collectionloghelper.lifecycle.OverlayRegistry;
 import com.collectionloghelper.lifecycle.PlayerLocationResolver;
 import com.collectionloghelper.lifecycle.SceneEventRouter;
@@ -157,6 +158,9 @@ public class CollectionLogHelperPlugin extends Plugin
 	@Inject
 	private VarbitChangeRouter varbitChangeRouter;
 
+	@Inject
+	private CollectionStateChangeHandler collectionStateChangeHandler;
+
 
 	private CollectionLogHelperPanel panel;
 	private NavigationButton navButton;
@@ -231,7 +235,16 @@ public class CollectionLogHelperPlugin extends Plugin
 		// Wire coordinator with references it needs from the plugin
 		guidance.getGuidanceCoordinator().setPluginInstance(this);
 		guidance.getGuidanceCoordinator().setPanel(panel);
-		guidance.getGuidanceCoordinator().setOnSequenceCompleteCallback(this::onSequenceComplete);
+		collectionStateChangeHandler.setCallbacks(
+			() ->
+			{
+				pendingPanelRebuild = true;
+				rankedSourcesDirty = true;
+			},
+			this::getRankedSources,
+			this::activateGuidance);
+		guidance.getGuidanceCoordinator().setOnSequenceCompleteCallback(
+			collectionStateChangeHandler::handleSequenceComplete);
 		// Wire coordinator into resolver (post-construction, avoids circular injection)
 		guidance.getRequiredItemResolver().setCoordinator(guidance.getGuidanceCoordinator());
 
@@ -283,7 +296,7 @@ public class CollectionLogHelperPlugin extends Plugin
 				travelCapabilities.refreshQuestState();
 				travelCapabilities.refreshVarbits();
 				sync.getSyncStateCoordinator().setLastObtainedCount(data.getCollectionState().getTotalObtained());
-				rebuildSourcesWithMissingItems();
+				sourcesWithMissingItems = collectionStateChangeHandler.rebuildSourcesWithMissingItems();
 				if (panel != null)
 				{
 					panel.rebuild();
@@ -304,7 +317,7 @@ public class CollectionLogHelperPlugin extends Plugin
 				pendingRequirementsRefresh = true;
 				pendingTravelVarbitRefresh = true;
 			},
-			this::rebuildSourcesWithMissingItems,
+			() -> sourcesWithMissingItems = collectionStateChangeHandler.rebuildSourcesWithMissingItems(),
 			() ->
 			{
 				pendingPanelRebuild = true;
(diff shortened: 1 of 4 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#540
**Files changed**: 4  (+518 / -127)
**Reviewer focus**: callback wiring, behavior parity at the 3 `rebuildSourcesWithMissingItems` call sites, the `OnSequenceCompletePerItemTest` rewrite scope, Mockito final-class workaround signal, and constructor visibility.

### Findings

#### CRITICAL (0)
None.

#### HIGH (0)
None.

#### MEDIUM (2)

- **`OnSequenceCompletePerItemTest` no longer exercises the full B4.4 chain end-to-end.** The pre-patch test verified `coordinator.activateGuidance(source, location, bestItemId)` was called through the production `plugin::onSequenceComplete -> plugin::activateGuidance(source, targetItemId)` chain. The rewrite verifies only that the handler hands `(source, bestItemId)` to its `activateGuidanceCallback`. The intermediate hop `plugin.activateGuidance(CollectionLogSource, Integer)` is not covered by any other test (the existing `ActivateGuidanceClientThreadTest` only exercises the 1-arg overload for client-thread routing). The 2-arg overload predates this PR and is unchanged, so this is a coverage shrinkage rather than a regression — but worth tracking. Consider a 2-line test in `ActivateGuidanceClientThreadTest` asserting the 2-arg overload forwards `targetItemId` unchanged to `coordinator.activateGuidance(source, location, targetItemId)`.

- **Handler keeps `Client` field but only uses it as a passthrough to `EfficiencyCalculator.exportEfficiencyList`.** `client` is held purely to relay into `exportEfficiencyList(file, client)`. If `EfficiencyCalculator` is also a `@Singleton` with its own `Client`, this dependency is redundant. Low-impact, but flagging since the handler's constructor list is wider than its actual collaborator surface.

#### LOW (2)

- **Public constructor on `CollectionStateChangeHandler` vs package-private on `VarbitChangeRouter`** (called out in PR description). The reason given — `OnSequenceCompletePerItemTest` lives in `com.collectionloghelper` rather than `com.collectionloghelper.lifecycle` — is reasonable. Counter-option: move the test to `com.collectionloghelper.lifecycle` (it tests a lifecycle component) and tighten visibility to package-private for consistency with Wave 12. Not blocking; either choice is defensible.

- **Three-arg `setCallbacks` boundary is bordering on a constructor smell.** Three lambdas wired post-construction to dodge a Guice cycle is the same pattern as `GuidanceOverlayCoordinator.setPluginInstance` / `setPanel`. Acceptable for now, but as the plugin shrinks past Wave 14 it would be cleaner to flip the dependency: have the plugin inject the handler with a small `PluginStateAccessor` interface implemented inline, so handler unit tests can mock the accessor rather than wire three lambdas. Tracking suggestion only.

### Verification ran
- [x] CI: both `Build` and `CI` workflow runs green (run 26062295374 / 26062295378, 2026-05-18 21:46Z).
- [x] PR description claims 1267 tests pass; CI confirms.
- [x] Verified 3 call sites for `rebuildSourcesWithMissingItems()` in the post-patch `Plugin.java` (lines 299, 320, 394) all assign back into `sourcesWithMissingItems` — behavior parity preserved.
- [x] Verified `sourcesWithMissingItems.clear()` semantics on `shutDown` (line 356) and login-screen reset (line 431) still operate on the plugin-owned field; handler is stateless w.r.t. that set.
- [x] Verified callbacks are wired once in `startUp()` and stored as handler fields (`onSequenceCompleteFlags`, `rankedSourcesSupplier`, `activateGuidanceCallback`) — no per-event allocation.
- [x] Verified `Plugin.java` LOC: 678 (matches PR description).
- [x] Verified handler LOC: 202 (well within file-size budget); methods all <50 LOC.
- [x] Verified `handleSequenceComplete` allocation-free guard-fail path: `autoAdvanceGuidance == false` returns after flag callback with zero allocations (matches the #527 / #523 hot-path discipline cited in PR body).

### Recommendation
**SHIP** — no CRITICAL or HIGH issues. The extraction is clean, behavior parity holds at all three `rebuildSourcesWithMissingItems` call sites, the callback wiring is per-startup not per-event, and the new handler test fills in the additional edge cases (locked-source skip, null-supplier-result, null-callbacks-don't-NPE, export-disabled no-op) that the prior reflection-based test did not exercise. The B4.4 contract is preserved at the handler boundary; the small coverage gap on `plugin.activateGuidance(source, targetItemId)` is pre-existing and worth a 2-line follow-up but not blocking.

### Notes
- Wave 14 candidate (`onGameTick` ~85 LOC, lines 497-577) is well-scoped in the PR description and looks tractable.
- The handler's `handleSequenceComplete` null-callback safety (`if (onSequenceCompleteFlags != null)` etc.) is defensive against test misuse; in production all three are always wired in `startUp`. The defense is cheap and is exercised by `handleSequenceComplete_nullCallbacks_doNotNpe`, so worth keeping.
- The Mockito final-class workaround using real `CollectionLogSource` instances via `minimalSource(...)` does not reduce signal — the assertions key off identity (`eq(source)`) and a primitive field (`bestItem.getItemId()`), both of which work with real instances.

PR #543, opened 2026-05-19
fix: add @ImplementedBy for player-state interfaces (plugin-hub blocker)

cha-ndler:
## Summary

Live smoke test on `./gradlew run` revealed a `PluginInstantiationException` that has existed since PRs #459/#461: Guice cannot resolve `DiaryTierState`, `EquippedItemState`, `PohTeleportInventory`, or `SkillCapePerkState`, all injected into `PlayerCapabilityDebugOverlay` via `OverlayRegistry`. One leaf failure kills the whole plugin — the Plugin Hub build would not load.

### Root cause

The four `Impl` classes carry `@Singleton` (enough for Guice to JIT-bind the **concrete** class), but the overlay injects the **interface**. No `@ImplementedBy` on the interfaces, no `bind(I.class).to(Impl.class)` anywhere in `src/main/java`. Existing unit tests pass because each test constructs the `Impl` directly or binds it by hand; the real injector has no such hint.

### Fix

- Add `@ImplementedBy(<Impl>.class)` to each of the four interfaces. Smallest possible diff — zero changes to impl bodies, no new modules.
- New regression test `CollectionLogHelperPluginInjectorTest` spins up a minimal Guice injector that binds only the leaf dependencies (`Client`, `CollectionLogHelperConfig`) and asks for each interface. If a future change drops an `@ImplementedBy` or adds another unbound interface dep, the test fails at CI time — not at Plugin Hub time.

### Test signal verified

Temporarily removed `@ImplementedBy` from `DiaryTierState` locally:
- `diaryTierState_resolvesToImpl` → `com.google.inject.ConfigurationException`
- `allPlayerStateInterfaces_resolveTogether` → `com.google.inject.ConfigurationException`

Restored the annotation and confirmed all 5 cases pass.

## Test plan

- [x] `./gradlew test` passes (including the new regression test)
- [x] New regression test fails on the unfixed code (proven by reverting one annotation)
- [ ] Reviewer re-runs `./gradlew run` to confirm the live client now starts the plugin without `PluginInstantiationException`

### DIFF
--- a/src/main/java/com/collectionloghelper/player/DiaryTierState.java
+++ b/src/main/java/com/collectionloghelper/player/DiaryTierState.java
@@ -24,6 +24,8 @@
  */
 package com.collectionloghelper.player;
 
+import com.google.inject.ImplementedBy;
+
 /**
  * Read-only view of a player's Achievement Diary completion state.
  * <p>
@@ -38,6 +40,7 @@ package com.collectionloghelper.player;
  * }
  * }</pre>
  */
+@ImplementedBy(DiaryTierStateImpl.class)
 public interface DiaryTierState
 {
 	/**
--- a/src/main/java/com/collectionloghelper/player/EquippedItemState.java
+++ b/src/main/java/com/collectionloghelper/player/EquippedItemState.java
@@ -24,6 +24,7 @@
  */
 package com.collectionloghelper.player;
 
+import com.google.inject.ImplementedBy;
 import java.util.Set;
 
 /**
@@ -46,6 +47,7 @@ import java.util.Set;
  * }
  * }</pre>
  */
+@ImplementedBy(EquippedItemStateImpl.class)
 public interface EquippedItemState
 {
 	/**
--- a/src/main/java/com/collectionloghelper/player/PohTeleportInventory.java
+++ b/src/main/java/com/collectionloghelper/player/PohTeleportInventory.java
@@ -24,6 +24,7 @@
  */
 package com.collectionloghelper.player;
 
+import com.google.inject.ImplementedBy;
 import java.util.Set;
 
 /**
@@ -53,6 +54,7 @@ import java.util.Set;
  * <p>Wiring into Tier B guidance branching is deferred to milestone C6.
  * This interface is intentionally standalone for C1.
  */
+@ImplementedBy(PohTeleportInventoryImpl.class)
 public interface PohTeleportInventory
 {
 	/**
--- a/src/main/java/com/collectionloghelper/player/SkillCapePerkState.java
+++ b/src/main/java/com/collectionloghelper/player/SkillCapePerkState.java
@@ -24,6 +24,8 @@
  */
 package com.collectionloghelper.player;
 
+import com.google.inject.ImplementedBy;
+
 /**
  * Read-only view of which skill-cape perks are currently available to
  * the logged-in player.
@@ -52,6 +54,7 @@ package com.collectionloghelper.player;
  * }
  * }</pre>
  */
+@ImplementedBy(SkillCapePerkStateImpl.class)
 public interface SkillCapePerkState
 {
 	/**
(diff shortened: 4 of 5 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#543
**Files changed**: 5 (+164/-0)
**Reviewer focus**: correctness of `@ImplementedBy` placement, import correctness, regression-test signal strength, preventive scan for other unbound interface deps in the object graph.

### Findings

#### CRITICAL (0)
None.

#### HIGH (0)
None.

#### MEDIUM (1)
- **Regression-test generality is enumerative, not structural.** The new `CollectionLogHelperPluginInjectorTest` resolves the four known interfaces by name. If a future change introduces a *fifth* interface dependency without an `@ImplementedBy` (the exact failure class this PR is shielding against), the test will not detect it — Plugin Hub will. A stronger shield would `Guice.createInjector(...)` a module that mirrors the real plugin wiring (or at minimum, walk `@Inject`-annotated constructors via reflection and assert each parameter type resolves). Acceptable as-is for the blocker fix, but worth tracking as a follow-up; otherwise we are one new interface away from a repeat of this incident.

#### LOW (2)
- The test's class-level Javadoc references "the four interfaces" — accurate today but will go stale if the list grows. Consider phrasing as "the known set of player-state interfaces (enumerated in the @Test methods below)."
- `PluginInstantiationException` is mentioned in the PR body but the test asserts only `ConfigurationException` (the underlying Guice cause). Worth a one-line comment in the test noting the wrapping relationship, so future readers don't grep for the wrong exception type.

### Verification ran
- [x] `./gradlew test` — all tests green including the new `CollectionLogHelperPluginInjectorTest` (5/5)
- [x] `./gradlew test --tests CollectionLogHelperPluginInjectorTest` — 5/5 green in isolation
- [x] All four `@ImplementedBy` annotations placed on the interface, pointing at the correct `*Impl` class (verified per-interface)
- [x] Import is `com.google.inject.ImplementedBy` in all four files (not `javax.inject` / `jakarta.inject`)
- [x] All four `*Impl` classes carry `@Singleton` and have `@Inject` constructors taking `Client` (and `CollectionLogHelperConfig` for `PohTeleportInventoryImpl`) — matches the test's minimal module bindings
- [x] Test module does **not** `bind(<Interface>).to(<Impl>)` — resolution genuinely flows through `@ImplementedBy`, so the shield is real
- [x] `allPlayerStateInterfaces_resolveTogether` resolves all four against a single injector (catches partial-binding regressions)
- [x] Preventive scan of `src/main/java/com/collectionloghelper/` for other `public interface` declarations: only the four touched here are injected by interface type in the production object graph. `QuestProgressState` exists as an interface but `PlayerCapabilityDebugOverlay` injects the concrete `PlayerQuestProgressState` directly, so JIT binding suffices. `DynamicTargetEvaluator` and `BossGuidance` are populated into registries via concrete impls, not Guice-resolved by interface. `PanelModeController` / `PanelShellContext` are not `@Inject`-ed by interface. No other latent offenders detected.
- [x] PR is `MERGEABLE`; both Build and CI workflows green on latest commit

### Recommendation
- **SHIP** — no CRITICAL/HIGH issues. MEDIUM is a follow-up improvement, not a blocker. This unblocks Plugin Hub resubmission and the smallest-possible-diff approach is correct for an urgent runtime fix.

### Notes
- Smallest-possible-diff is the right call here: zero changes to `Impl` bodies, no new Guice module, four annotations and a test. Reversible and low-risk.
- The regression test's value is real even with the enumerative caveat above — it locks down the four interfaces against future drift (e.g., someone "cleaning up unused imports" and removing an `@ImplementedBy`).
- Consider opening a follow-up issue to harden the shield: a structural test that introspects every `@Inject` constructor in `src/main/java` and asserts each parameter type can be resolved by Guice against a real plugin-module mirror. That would catch the bug class generically, not just the four known instances.

PR #541, opened 2026-05-19
refactor: extract GameTickOrchestrator from Plugin (#503)

cha-ndler:
## Summary

Extracts the per-tick coalescer body from `CollectionLogHelperPlugin.onGameTick` into a new `GameTickOrchestrator` lifecycle collaborator. Plugin retains the `@Subscribe onGameTick` annotation and delegates to a single `gameTickOrchestrator.tick()` call. Part of #503 (Wave 14).

## LOC delta

| File | Before | After | Delta |
| --- | --- | --- | --- |
| `CollectionLogHelperPlugin.java` | 678 | 628 | -50 |
| `GameTickOrchestrator.java` (new) | -- | 272 | +272 |
| `GameTickOrchestratorTest.java` (new) | -- | 482 | +482 |

The plugin lands at **628 LOC** -- still above the 600 architectural target, but the residual is dominated by `startUp` glue (panel construction, callback wiring across 7 collaborators) and `shutDown` teardown. Further reduction would require splitting `startUp`/`shutDown` themselves, which is a separate seam (likely an `OnStartupRoutine`/`OnShutdownRoutine` extraction worth its own wave).

## Hot-path allocation analysis

`onGameTick` is the hottest path in the plugin (~every 600ms). The reviewer should focus here:

1. **Plugin keeps ownership of every per-tick state flag** (`pendingRequirementsRefresh`, `pendingTravelVarbitRefresh`, `slayerRefreshPending`, `pendingPanelRebuild`, `rankedSourcesDirty`). The orchestrator never touches these fields directly.

2. **All callbacks are method references**, not lambdas:
   - `this::pollRequirementsRefresh`, `this::pollTravelVarbitRefresh`, `this::pollSlayerRefresh`, `this::pollPanelRebuild` (poll-and-clear)
   - `this::markPanelRebuildAndRankedDirty`, `() -> rankedSourcesDirty = true` (dirty setters)
   - `() -> panel` (panel accessor)

   The JVM hoists method references to constants; the two trivial setter lambdas are also captured once during `startUp` and stored on the orchestrator's fields. No per-tick lambda allocation.

3. **No-op tick allocation profile** (covered by `GameTickOrchestratorTest.noOpTick_doesNotTouchAnyMutator`): when every flag is false and the player is absent, the tick body executes:
   - 4 primitive boolean polls (zero allocation)
   - 1 non-null reference read on `pluginDataManager.getCharacterDir()` (zero allocation)
[...]

PR #551, opened 2026-05-19
test: harden PanelRebuildOrchestratorTest against cross-test Swing state contamination

cha-ndler:
## Summary

Preventive hardening for the headless-Linux scrollbar flake on `PanelRebuildOrchestratorTest`.

#536 fixed the in-class flake (JDK 17 and JDK 11) by swapping the vertical scrollbar's `BoundedRangeModel` for a `PinnedRangeModel` whose range-narrowing setters become no-ops once pinned. The reviewer on a follow-up PR observed the same flake re-emerge during a *full-suite* test run on `captureRecordsScrollPositionFromEnclosingScrollPane`, while it remained green when the class was run in isolation. That points at cross-test contamination: sibling-test EDT events landing on the bar between the initial `JScrollPane` layout pass in `setUp()` and the `pinScrollBounds(...)` call.

## Root cause hypothesis

In the previous `setUp()`:

1. `scrollPane.setVerticalScrollBar(new JScrollBar(...))` -- bar joins scrollpane with the *default* model (max=100, extent=10).
2. `scrollPane.setSize(...)`, `scrollPane.doLayout()`, `viewport.doLayout()` -- a headless-CI layout pass can call `bar.setMaximum(...)` / `setValues(...)` during this window.
3. `pinScrollBounds(bar)` -- swaps in the `PinnedRangeModel` *after* layout.

That ordering leaves a brief window (between step 1 and step 3) during which any pending EDT event from a previously-run test class could narrow the bar's model. Once the next assertion reads `bar.getValue()`, the clamp has already happened.

## Fix (Options A + B + D combined)

1. **Pre-pin the model before the bar joins the scrollpane.** Build a `JScrollBar`, install a `PinnedRangeModel` on it, write+pin the canonical range (`value=0, extent=400, min=0, max=5000`), *then* call `scrollPane.setVerticalScrollBar(bar)`. This eliminates the layout-pass narrowing window entirely. The existing `pinScrollBounds(...)` call is kept as an idempotent re-pin.

2. **New `@After tearDown()`** that:
   - Swaps the pinned scrollbar out for a fresh `JScrollBar` so any deferred listeners on the pinned bar see a no-op recipient.
   - Drains the AWT EventQueue via `EventQueue.invokeAndWait(() -> {})` so `invokeLater()` callables scheduled by `deferScrollRestore*` tests cannot fire during the next test's `setUp()`.

3. **`@FixMethodOrder(MethodSorters.NAME_ASCENDING)`** so any future contamination-style failure is reproducible rather than dependent on JVM hash-ordering of test methods.

## Reproduction history

- Two master CI failures recorded before #536 landed: runs 26054950581 (`deferScrollRestoreSchedulesValueOnVerticalScrollBar`) and 26056163526 (`deferScrollRestoreIsNoOpWhenScrollPositionIsZero`), both on `headSha` 8a50f46d / 58b1314b.
- Post-#536, the reviewer on PR #541 saw the flake re-emerge in their full-suite run on `captureRecordsScrollPositionFromEnclosingScrollPane`, passing in isolation. Master CI itself has remained green since #536, so this is preventive hardening rather than a reproducible local fix.
- Cannot reproduce locally on Windows JDK 17 (5/5 green on `./gradlew test --rerun-tasks` against the pre-fix code).

## Test plan

- [x] `./gradlew test --tests "com.collectionloghelper.ui.PanelRebuildOrchestratorTest" --rerun-tasks` -- green
- [x] `./gradlew test --rerun-tasks` (full suite) x 5 consecutive runs on Windows JDK 17 -- 5/5 green
- [ ] CI Build (Linux JDK 11) -- green
- [ ] CI workflow (Linux JDK 17) -- green

### DIFF
--- a/src/test/java/com/collectionloghelper/ui/PanelRebuildOrchestratorTest.java
+++ b/src/test/java/com/collectionloghelper/ui/PanelRebuildOrchestratorTest.java
@@ -30,6 +30,7 @@ import com.collectionloghelper.data.PlayerCollectionState;
 import com.collectionloghelper.data.RequirementsChecker;
 import java.awt.BorderLayout;
 import java.awt.Dimension;
+import java.awt.EventQueue;
 import java.util.Collections;
 import java.util.Set;
 import javax.swing.BoxLayout;
@@ -40,9 +41,12 @@ import javax.swing.JScrollBar;
 import javax.swing.JScrollPane;
 import javax.swing.SwingUtilities;
 import net.runelite.client.game.ItemManager;
+import org.junit.After;
 import org.junit.Before;
+import org.junit.FixMethodOrder;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.junit.runners.MethodSorters;
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnitRunner;
 
@@ -60,6 +64,7 @@ import static org.mockito.Mockito.when;
  * of issue #503 god-class splits.
  */
 @RunWith(MockitoJUnitRunner.class)
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
 public class PanelRebuildOrchestratorTest
 {
 	@Mock
@@ -77,7 +82,7 @@ public class PanelRebuildOrchestratorTest
 	private PanelRebuildOrchestrator orchestrator;
 
 	@Before
-	public void setUp()
+	public void setUp() throws Exception
 	{
 		when(collectionState.getCategoryCount(any())).thenReturn(0);
 		when(collectionState.getCategoryMax(any())).thenReturn(10);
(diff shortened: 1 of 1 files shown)


cha-ndler:
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#551
**Files changed**: 1 (+57/-14, test-only)
**Reviewer focus**: correctness of the pre-pin ordering, soundness of the `@After` EDT drain, behaviour parity of the existing test assertions, and absence of any production-code regression risk.

### Findings

#### CRITICAL (0)
- None.

#### HIGH (0)
- None.

#### MEDIUM (1)

- **`tearDown()` swaps in `new JScrollBar(JScrollBar.VERTICAL)` whose model is the JDK default (`min=0, max=100, extent=10, value=0`).** This is intentional (you want the pinned bar gone so its listeners stop receiving events), but the resulting fresh bar will receive any in-flight layout events with the narrow range. Since the next test's `setUp()` allocates a fresh `JScrollPane` and a fresh pre-pinned bar, this is harmless -- the post-`tearDown` bar is GC-eligible immediately. Flagging only because the comment "fresh, non-pinned" might read as "ready for re-use"; it is not. Consider rewording to `// discard the pinned bar; the next setUp() rebuilds everything` or similar. Non-blocking.

#### LOW (2)

- **`@FixMethodOrder(MethodSorters.NAME_ASCENDING)`** locks in this insertion order: `captureIgnoresNonCategoryChildren`, `captureRecordsOnlyCurrentlyExpandedCategoryNames`, `captureRecordsScrollPositionFromEnclosingScrollPane`, `captureReturnsZeroScrollAndEmptySetWhenNothingExpanded`, `deferScrollRestoreIsNoOpWhenPanelNotInScrollPane`, `deferScrollRestoreIsNoOpWhenScrollPositionIsZero`, `deferScrollRestoreSchedulesValueOnVerticalScrollBar`, `restoreExpandedDoesNotReExpandRemovedCategory`, `restoreExpandedIsNoOpWhenSnapshotIsEmpty`, `restoreExpandedReExpandsMatchingCategoriesAfterRebuild`, `snapshotExpandedCategoriesIsImmutable`. No cross-test data dependencies, so correctness is unaffected. The trade-off is that future renames can silently change the run order; an inline comment near the annotation would help future readers.
- **`@After public void tearDown() throws Exception`** declares `throws Exception` to accommodate `EventQueue.invokeAndWait`'s checked `InterruptedException` and `InvocationTargetException`. JUnit 4 will surface either as a test-class teardown failure, which is the correct behaviour. No action.

### Pre-pin ordering verification

Walked the new `setUp()` against the pre-fix flake hypothesis:

1. `new JScrollBar(JScrollBar.VERTICAL)` -- bar is born with the JDK `DefaultBoundedRangeModel(value=0, extent=10, min=0, max=100)`. Bar is not attached to any container; no listeners are wired beyond its own `BoundedRangeModelHandler`.
2. `new PinnedRangeModel()` -- the subclass field `pinned` defaults to `false`. Inherited `DefaultBoundedRangeModel` state matches the JDK default.
3. `preModel.setRangeProperties(0, 400, 0, 5000, false)` -- because `pinned == false` this hits the `super.setRangeProperties(...)` branch and writes the four ranges directly. Confirmed by re-reading the override at lines 222-238 of the new file. After this call the model is `(value=0, extent=400, min=0, max=5000)`.
4. `preModel.pin()` -- flips `pinned = true`. From this point onward, all `setExtent / setMinimum / setMaximum / setRangeProperties` calls on this model become no-ops (or clamp-and-write for value).
5. `bar.setModel(preModel)` -- the bar's `BoundedRangeModelHandler` detaches from the JDK default model and attaches to `preModel`. Any internal `BasicScrollBarUI` listeners follow the swap. **Crucial: the bar's model is fully pinned before it ever touches a `JScrollPane`.**
6. `scrollPane.setVerticalScrollBar(bar)` -- the scrollpane's `ScrollPaneLayout` may now register a `ChangeListener` on `bar.getModel()` (which is `preModel`). Any subsequent `scrollPane.doLayout()` or `viewport.doLayout()` that would have called `bar.setMaximum(viewSize.height)` (the suspected pre-fix flake vector) is silently dropped by `preModel`.

The window that the previous code left open -- between `scrollPane.setVerticalScrollBar(new JScrollBar(...))` and `pinScrollBounds(...)` in the old setUp -- is now zero-width. **The fix is correct.**

### `@After` drain verification

Walked the new `tearDown()`:

1. `scrollPane.setVerticalScrollBar(new JScrollBar(JScrollBar.VERTICAL))` -- the pinned bar is detached. Any listeners that the scrollpane registered on the pinned bar's model are also detached (via the model-swap inside `setVerticalScrollBar`). Confirmed by reading the JDK source for `JScrollPane.setVerticalScrollBar(JScrollBar newScrollBar)`: it calls `add(newScrollBar)` after removing the old one, which drops the old listener.
2. `EventQueue.invokeAndWait(() -> {})` -- this is called from the JUnit worker thread (NOT the EDT), which is the correct invocation context. From the EDT it would throw `IllegalThreadStateException`; JUnit 4 does not pin tests to the EDT and this test class does not wrap any of its methods in `SwingUtilities.invokeAndWait`. The no-op runnable runs after every other event already queued on the EDT, so any `invokeLater` scheduled by `deferScrollRestore` (which the three `deferScrollRestoreX` tests exercise) has fully drained before the next test starts.

Two micro-edge cases worth recording but not blocking:

- If a future test ever wraps its body in `SwingUtilities.invokeAndWait(...)` to run on the EDT, the `@After` will run on the EDT too and `EventQueue.invokeAndWait` will throw. None of the existing tests do this. A guard `if (!EventQueue.isDispatchThread()) { ... }` would future-proof this, but is YAGNI for the current scope.
- `EventQueue.invokeAndWait` propagates `InterruptedException`; under test cancellation the teardown will exit early. Acceptable.

### Behaviour parity (all 11 test methods)

- `captureReturnsZeroScrollAndEmptySetWhenNothingExpanded` -- reads bar value via `orchestrator.capture()`, expects 0. With the pre-pinned model at value=0, this is unchanged.
- `captureRecordsScrollPositionFromEnclosingScrollPane` -- the original flake target. Calls `pinScrollBounds(vBar())` (idempotent, model already pinned) then `vBar().setValue(250)`. `setValue` routes through `setRangeProperties` which the pinned override clamps to `max=5000-extent=4600`, so 250 sticks. Then `capture()` reads 250. Unchanged behaviour.
- `captureRecordsOnlyCurrentlyExpandedCategoryNames` / `captureIgnoresNonCategoryChildren` / `restoreExpanded*` / `snapshotExpandedCategoriesIsImmutable` -- do not touch the scrollbar; only the expanded-categories Set. Behaviour preserved.
- `deferScrollRestoreSchedulesValueOnVerticalScrollBar` -- sets value=400, captures, sets value=0, calls `deferScrollRestore(snapshot)`, drains EDT via `SwingUtilities.invokeAndWait(() -> {})`, asserts value=400. The `@After` drain is redundant with this test's internal drain but does no harm. Unchanged.
- `deferScrollRestoreIsNoOpWhenScrollPositionIsZero` -- captures at value=0, then re-pins and sets value=123, asserts no restore overwrites 123. The `@After` drains after the assertion; safe.
- `deferScrollRestoreIsNoOpWhenPanelNotInScrollPane` -- builds an orphan panel orchestrator; the class-level `scrollPane` is still the pinned one. `@After` drains after assertion; safe.

### Production-code surface

`git diff cca37bbb..HEAD -- src/main/` returns empty. **Zero production code touched.** This is a pure test-hardening PR. Risk surface is bounded to the test file.

### Cross-repo refs

Body mentions `#536`, `#541`, `#551`. All are within `cha-ndler/collection-log-helper`. PR is being posted on the same repo, so bare `#NNN` refs are correct here. No cross-repo violations.

### Verification ran

- [x] `./gradlew test --tests PanelRebuildOrchestratorTest --rerun-tasks` on Windows JDK 17 -- green
- [x] `./gradlew test --rerun-tasks` (full suite) x 5 consecutive runs on Windows JDK 17 -- 5/5 green
- [x] CI Build (Linux JDK 11) on run 26113910606 -- **pass** (51s)
- [x] CI workflow (Linux JDK 17) on run 26113910609 -- **pass** (1m2s)
- [x] `git diff cca37bbb..HEAD -- src/main` -- empty (no production change)
- [x] Confirmed `PinnedRangeModel.setRangeProperties` writes the full range correctly when called before `pin()` (the unpinned branch is hit because the field defaults to `false`).

### Recommendation

**SHIP** -- no CRITICAL or HIGH issues. The single MEDIUM (comment wording on the tearDown bar swap) is purely cosmetic. The fix correctly closes the pre-pin layout-pass window that #536 left open, the `@After` drain prevents inverse leakage into the next test's setUp, and the `@FixMethodOrder` makes future regressions reproducible. CI is green on both Linux JDK 11 and Linux JDK 17 on the first push.

### Notes

- The reproducer history is honest: this PR could not be reproduced locally on Windows (the headless-Linux specifics do not appear on Windows JDK 17 / Win 11), so this lands as preventive hardening. The fact that the reviewer on #541 observed the flake re-emerge in a full-suite run is the genuine motivating signal.
- If the flake ever re-appears post-merge despite all three hardenings, the next escalation is Option C from the original brief (fully self-contained test that injects a fresh `JScrollBar` per test method rather than reusing the scrollpane-managed one). That is materially larger surgery and is correctly deferred.
- The pin-before-attach pattern is generalisable; if any future Swing-test in this repo trips a similar headless-CI flake, the same idiom (build the model, write the canonical range, pin, install on the bar, then attach to the container) is the right starting point.

PR #580, opened 2026-05-21
fix(ui): wrap step description text in side panel to prevent mid-string truncation (#575)

### LINKED ISSUE #575: ui(panel): step description text truncated in side panel (in-game overlay renders correctly)
## Summary

Surfaced 2026-05-20 during #495 Pass 2 in-game validation on General Graardor. The active step's description in the **CLH side panel** is truncated mid-string. The same step renders correctly in the **in-game overlay**.

## What you should see vs. what the panel shows

| Where | Text |
|---|---|
| Data (\`drop_rates.json\`, step 4) | "Enter the Bandos boss room and kill General Graardor" |
| In-game overlay | "Enter the Bandos boss room and kill General Graardor" (correct) |
| **CLH side panel** | "Enter the Bandos boss room and Graardor" (truncated — missing "kill General") |

## Suspected cause

Side-panel step-text widget has a fixed width and clips mid-word instead of wrapping to a second line, using ellipsis at the end, or honoring the runelite panel's natural width.

Mid-string elision (not end-of-string) suggests this isn't a clean "fits/doesn't fit" decision but rather a font-render or layout-pass artifact — a Swing component sizing itself before the string is fully laid out, or a paint call truncating glyphs once a width budget is exceeded.

Worth checking:

- \`StepProgressView\` / \`StepProgressLabel\` widget in \`src/main/java/com/collectionloghelper/ui/\` (extracted as part of #503 god-class splits)
- \`CategorySummaryPanel\` layout pass

## Why this is a Plugin Hub blocker

User-facing visible bug. In-game overlay being correct is good, but the panel is the primary surface for guidance — a truncated kill step is misleading and unprofessional. Tagged \`tier-A\`.

## Repro

1. Activate guidance on **General Graardor** (or any source whose final step description exceeds ~40 chars)
2. Observe panel's blue active-step text
3. Note mid-string truncation

May also affect Kree'arra (longer kill-step with "mithril grapple" sub-instruction) and any other source with verbose step text.

cha-ndler:
## Summary

Fixes #575 — the active step's description was being truncated mid-string in the side panel (e.g. "Enter the Bandos boss room and Graardor" — missing "kill General").

## Root cause

The earlier #497 fix wrapped the step-progress label HTML body at `width: 220px` on the assumption that the plugin panel inner column was ~250px. The actual `PluginPanel.PANEL_WIDTH` is 225px:

| Layer | Width |
|---|---|
| `PluginPanel.PANEL_WIDTH` | 225 |
| Shell outer empty border (6+6) | -12 |
| `StepProgressView` line border (1+1) + empty border (6+6) | -14 |
| Vertical scrollbar (`PluginPanel.SCROLLBAR_WIDTH`) when content overflows | -17 |
| **Worst-case usable label width** | **~182** |

Asking Swing's HTML renderer to lay out at 220px when only ~182px is actually paintable causes glyphs past the clip region to be dropped silently with no end-of-string ellipsis. That matches the symptoms in #575 (mid-string elision, not a tidy "…").

## Fix

- Lowers `PANEL_TEXT_WRAP_WIDTH_PX` from 220 to **170**, which sits a few px below the worst-case usable width and matches the conservative width already used by `ItemDetailPanel`.
- HTML-escapes the step description before inlining it into the `<body>` so any future text containing `<`, `>`, `&` cannot corrupt the rendered output (defensive — current `drop_rates.json` text has no reserved chars).
- Adds three regression tests:
  - long description text appears verbatim in the rendered label
  - reserved HTML characters in the description are escaped
  - `escapeHtml` helper handles `null`, plain text, and all reserved chars

The in-game overlay path is unaffected — it does not use this label and was already rendering the description correctly.

## Files changed

- `src/main/java/com/collectionloghelper/ui/widget/StepProgressView.java`
- `src/test/java/com/collectionloghelper/ui/widget/StepProgressViewTest.java`

## Test plan

- [x] `./gradlew test --tests "com.collectionloghelper.ui.widget.StepProgressViewTest"` — passes
- [x] `./gradlew build` — green (compile + tests + JaCoCo + drop-rates lint)
- [ ] In-game smoke: activate guidance on General Graardor (sectioned source, long step-4 text) and verify the side-panel label reads "Enter the Bandos boss room and kill General Graardor" with multi-line wrap rather than mid-string elision
- [ ] In-game smoke: confirm the in-game overlay still renders the same description correctly (regression check that the fix didn't touch the overlay path)
- [ ] Visual check on a clue-tier source with long descriptions (the original #497 case — "Buy a sleek hairband from the Falador hairdresser") to confirm wrapping still works

Fixes #575.

### DIFF
--- a/src/main/java/com/collectionloghelper/ui/widget/StepProgressView.java
+++ b/src/main/java/com/collectionloghelper/ui/widget/StepProgressView.java
@@ -96,10 +96,27 @@ public class StepProgressView extends JPanel
 
 	/**
 	 * Inner-width budget (px) used to force Swing's JLabel HTML renderer to
-	 * word-wrap long step descriptions. Matches the usable column of the
-	 * default RuneLite plugin panel (~250px minus padding/scroll gutters).
+	 * word-wrap long step descriptions.
+	 *
+	 * <p>RuneLite's {@code PluginPanel.PANEL_WIDTH} is 225px. The shell panel
+	 * adds a 6px empty border on each side (12px total) and this widget adds a
+	 * 1px line border plus a 6px empty border on each side (~14px total). When
+	 * scrollable content is present the vertical scrollbar consumes another
+	 * 17px ({@code PluginPanel.SCROLLBAR_WIDTH}). That leaves roughly
+	 * {@code 225 - 12 - 14 - 17 = 182px} of usable label width in the worst
+	 * case.
+	 *
+	 * <p>An earlier #497 fix used 220px, which is wider than the worst-case
+	 * usable space. When Swing's HTML renderer is asked to lay out at a width
+	 * larger than the paint-clip region, glyphs past the clip get dropped
+	 * silently (no end-of-string ellipsis), producing the mid-string
+	 * truncation reported in #575 (e.g. "Enter the Bandos boss room and
+	 * Graardor" — missing "kill General"). 170px sits a few px below the
+	 * worst-case usable width so the JLabel never asks the layout engine for
+	 * more horizontal space than the panel can actually paint, matching the
+	 * conservative wrap width used by {@link com.collectionloghelper.ui.ItemDetailPanel}.
 	 */
-	private static final int PANEL_TEXT_WRAP_WIDTH_PX = 220;
+	private static final int PANEL_TEXT_WRAP_WIDTH_PX = 170;
 
 	private static final Color BG = new Color(25, 35, 55);
 	private static final Color SECTION_HEADER_FG = new Color(200, 200, 200);
@@ -481,9 +498,13 @@ public class StepProgressView extends JPanel
 			// past the panel's clip region — visible on clue tiers with long
 			// step text such as "Buy a sleek hairband from the Falador hairdresser",
 			// where the unwrapped label rendered outside the panel box (#483).
+			// HTML-escape the description so any future text containing reserved
+			// characters (<, >, &) does not corrupt the rendered output. The
+			// step number/total are plain integers and are safe to inline.
 			stepProgressLabel.setText(
 				"<html><body style='width:" + PANEL_TEXT_WRAP_WIDTH_PX + "px'>Step "
-					+ current + "/" + total + ": " + description + "</body></html>");
+					+ current + "/" + total + ": " + escapeHtml(description)
+					+ "</body></html>");
 			nextStepButton.setVisible(isManual);
 
 			if (groups.isEmpty())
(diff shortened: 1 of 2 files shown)


cha-ndler (commented):
## Self-review — severity-tagged

### [INFO] Root cause
The HTML wrap width was 220px but the worst-case paint-clip region in a RuneLite plugin panel (after panel border + widget border + scrollbar gutter) is ~182px. Swing's HTML renderer silently drops glyphs past the clip when the body width exceeds the paint region, producing mid-string elision instead of a clean end-of-string ellipsis. Lowering the wrap budget to 170px puts the layout inside the paintable column for every supported sidebar state.

### [INFO] Surface modified
- `src/main/java/com/collectionloghelper/ui/widget/StepProgressView.java` — only the `PANEL_TEXT_WRAP_WIDTH_PX` constant value (220 → 170), expanded javadoc, and a new `escapeHtml` helper invoked at the single `setText` site.
- `src/test/java/com/collectionloghelper/ui/widget/StepProgressViewTest.java` — three new tests + one new search helper.

No public API change. No new dependencies.

### [INFO] Why this surface
The truncation is observable only in the side panel's HTML label. The in-game overlay path was already correct, so it was not touched. The chip strips and section blocks compose at the layout level (BoxLayout sizing each chip), so they already wrapped — the bug was exclusive to the HTML-rendered description label.

Alternatives considered:
- Use `JTextArea` with `setLineWrap(true)` instead of HTML — works but loses the existing colour/font styling and requires re-wiring the parent BoxLayout sizing. More surface for the same outcome.
- Make the wrap width dynamic from `getWidth()` — fragile because the widget receives its width late in the layout pass, after `setText` has already computed preferred size. The original #497 chose a static budget for exactly this reason; the fix here keeps that decision and just corrects the magnitude.

### [INFO] Test coverage
- `showStep_longDescription_fullTextPresentInRenderedHtml` — pins the exact Bandos kill-step text from #575 as a verbatim substring of the rendered label
- `showStep_descriptionWithHtmlReservedChars_escapesThem` — pins the escape behaviour against future descriptions containing `<`, `>`, `&`
- `escapeHtml_handlesAllReservedCharsAndNull` — pure helper test (no EDT) for defence in depth

### [INFO] Build status
`./gradlew build` — green. Includes JaCoCo coverage gate and drop-rates lint.

### [INFO] Overlay path verified unaffected
`StepProgressView` is added only to the side panel via `PanelHeaderBuilder`. The `GuidanceOverlay` (`src/main/java/com/collectionloghelper/overlay/GuidanceOverlay.java`) renders step text through its own `Graphics2D` text path with `MAX_PANEL_WIDTH = 200` and was not touched.

**Verdict: SHIP.**

Issue #547, opened 2026-05-19
ux(panel): replace Skip/Reset/Sync text buttons in active guidance with icon controls + tooltips


cha-ndler:
## Summary

In the active guidance panel, the current action-button row reads:

```
[ Skip ] [ Reset ] [ Sync ]
```

These text labels obscure what each button actually does:

- **Skip** — advance to next step
- **Reset** — restart the **current task** from step 1 (does NOT stop guidance — discovered during live validation 2026-05-19)
- **Sync** — re-sync collection log state

The "Reset doesn't stop guidance" behavior is surprising — users intuitively expect Reset to terminate the current task, but it actually re-enters step 1 of the same task. To actually stop guidance the user must use a separate "Deactivate" action (currently elsewhere in the UI / via a different mechanism).

## Suggested fix

Replace the three text buttons with a familiar icon control row (like a music player), each with a hover tooltip explaining the action in plain words:

| Icon | Action | Tooltip |
|---|---|---|
| `<` (back chevron) | Previous step (if applicable) | "Go back one step" |
| `>` (forward chevron / skip-next) | Advance to next step | "Skip to next step" |
| `⏹` (stop square) | Deactivate guidance | "Stop guidance — clear all overlays" |
| `↻` (refresh / circular arrow) | Restart current task from step 1 | "Restart this task from the first step" |
| `↻-with-cloud` or `⭯` | Sync from collection-log | "Sync collection-log state" |

The icons should:
- Be visually familiar (music-player metaphor)
- Have hover tooltips with full plain-English descriptions
- Be sized comfortably for click targets (≥ 24x24)

## Why this matters

- Eliminates the "Reset doesn't actually stop guidance" surprise
- More discoverable than text-button labels
- Aligns with established patterns from media controls / RuneLite's own icon-driven UI elsewhere

## Discovered during

2026-05-19 interactive validation session, Phase D (Guide Me activation lifecycle). User clicked Reset expecting deactivation; observed task restart instead. Suggested icon UX in real-time.

## Scope

- Pure UI change in `CollectionLogHelperPanel` (or wherever the guidance panel's button row is built)
- Probably ~30-60 LOC including icon assets if not already available
- Add tooltips via Swing's `setToolTipText`

## Related

- Original button row predates the 38-PR refactor; this is a separate UX improvement track from #503 architectural work
- Could potentially fold into the same PR as #545 (SLAYER count fix) since both are panel UX, but probably cleaner as standalone

## Acceptance

[...]

PR #581, opened 2026-05-21
fix(ui): sync step-control STOP icon with source-level Stop Guidance button (#576)

### LINKED ISSUE #576: ui(guidance): step-control STOP button doesn't sync state with source-level Stop Guidance button
## Summary

Surfaced 2026-05-20 during #495 Pass 2 in-game validation on General Graardor. The red STOP icon in the step-instruction control bar (the one that ended up at the bottom of the active-step widget after the #547 icon-controls work) does NOT sync state with the source-level "Stop Guidance" button at the bottom of the collection log item that initiated guidance.

## Repro

1. Activate guidance on any source (General Graardor used here) via the source-level "Guide Me" button
2. Both the step-control STOP icon (red square in the step widget) AND the source-level "Stop Guidance" button at the bottom of the source row are now in "active guidance" state
3. Click the **red STOP icon** in the step controls
4. Step-by-step guidance stops as expected
5. **BUT** the source-level button at the bottom still says "Stop Guidance" — it doesn't revert to "Guide Me"

## Expected

Both controls should be views of the same underlying guidance-active state. Clicking either should:
- (a) Stop guidance
- (b) Flip BOTH controls back to their inactive state ("Guide Me" / play icon respectively)

## Why this is a Plugin Hub blocker

The two controls now disagree about whether guidance is active. From the user's perspective the panel says "Stop Guidance" (which implies "active, click to stop") but step-by-step events have stopped — a confusing state desync that erodes trust in the panel.

Tagged \`tier-A\`.

## Suspected cause

The #547 icon-controls extraction introduced a separate event path for the new step-control buttons that doesn't fire the same source-level state-update event as the existing "Stop Guidance" button. Likely a missing event dispatch in:

- \`StepControlIcons.java\` (introduced #547)
- Or whichever component owns the source-level button state machine

The fix is probably a one-line "also dispatch source-level deactivate event when STOP icon clicked" — or refactor so both buttons subscribe to the same canonical guidance-state observable.

## Adjacent question

Worth confirming the inverse case too: does clicking source-level "Stop Guidance" correctly flip the step-control icons? (Likely yes since that's the older path, but verify.)

cha-ndler:
## Summary

Fixes #576. The step-control STOP icon (the red square in the
step-instruction control bar introduced in #547) used to deactivate
guidance without flipping the source-level Guide Me / Stop Guidance
button on the active item-detail card. Both controls drive the same
underlying guidance-active state, so they need to render in lockstep.

## Root cause

The deactivate fan-out chain is:

`StepProgressView` stop callback -> `CollectionLogHelperPlugin.deactivateGuidance` ->
`GuidanceOverlayCoordinator.deactivateGuidance` -> `OverlayDeactivator.deactivate` ->
`panel.setGuidanceState(false, null, null)`

`setGuidanceState` updated the panel's own `guidanceActive` field and
called `quickGuidePanelView.syncGuidanceState(false, null)`, but never
reached the `ItemDetailPanel` currently rendered in the detail card.
`ItemDetailPanel` only updated its button text/background from inside its
own action listener (lines 213-228), so the only path that kept it in
sync was a click on its own Stop Guidance button.

## Fix

- Added `ItemDetailPanel.syncGuidanceState(active, guidedSource)` —
  the same shape as the existing `QuickGuidePanelView.syncGuidanceState`.
- Added `DetailViewBuilder.syncGuidanceState(active, guidedSource)` that
  forwards to the most-recently-rendered detail panel (no-op when none).
- `CollectionLogHelperPanel.setGuidanceState` now fans out to both
  `quickGuidePanelView.syncGuidanceState` and
  `detailViewBuilder.syncGuidanceState`, so every surface stays in sync
  no matter which control initiated the activate/deactivate.

## Test plan

- [x] `./gradlew build` green (Checkstyle, all tests, JaCoCo 80% gate).
- [x] New unit tests in `DetailViewBuilderTest`:
  - `syncGuidanceStateFlipsStopGuidanceButtonBackToGuideMe` — STOP icon path.
  - `syncGuidanceStateFlipsGuideMeButtonToStopGuidanceWhenActivatedElsewhere` — inverse direction (activate from elsewhere).
  - `syncGuidanceStateBeforePopulateIsNoOp` — guards the never-entered-detail-view path.
- [ ] Manual: click the source-level Guide Me button -> step-control row appears with active state; click the red STOP icon -> source-level button reverts to Guide Me without leaving detail view.
- [ ] Manual: with detail view open and guidance active, click the Quick Guide Stop Guidance button -> detail-view button reverts too.

### DIFF
--- a/src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java
+++ b/src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java
@@ -524,6 +524,11 @@ public class CollectionLogHelperPanel extends PluginPanel implements PanelShellC
 			guidanceBannerView.hideGuidance();
 		}
 		quickGuidePanelView.syncGuidanceState(active, source);
+		// #576 — sync the source-level Guide Me / Stop Guidance button on the
+		// active item-detail card. Without this fan-out, deactivating guidance
+		// from the step-control STOP icon would update the Quick Guide button
+		// but leave the detail-view button stuck in "Stop Guidance" mode.
+		detailViewBuilder.syncGuidanceState(active, source);
 	}
 
 	public void showClueGuidance(CollectionLogSource source)
--- a/src/main/java/com/collectionloghelper/ui/DetailViewBuilder.java
+++ b/src/main/java/com/collectionloghelper/ui/DetailViewBuilder.java
@@ -57,6 +57,17 @@ public class DetailViewBuilder
 	private final BiConsumer<CollectionLogSource, Integer> guidanceActivator;
 	private final Runnable guidanceDeactivator;
 
+	/**
+	 * Most-recently-rendered detail panel. Held so that
+	 * {@link #syncGuidanceState(boolean, CollectionLogSource)} can reach the
+	 * currently-visible Guide Me / Stop Guidance button when guidance is
+	 * deactivated from a different surface (e.g., the step-control STOP icon
+	 * in {@code StepProgressView}). Mirrors the {@code lastGuideButton} held by
+	 * {@link com.collectionloghelper.ui.widget.QuickGuidePanelView}. See
+	 * cha-ndler/collection-log-helper#576.
+	 */
+	private ItemDetailPanel lastDetail;
+
 	public DetailViewBuilder(
 		PlayerCollectionState collectionState,
 		RequirementsChecker requirementsChecker,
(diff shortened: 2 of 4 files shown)


cha-ndler (commented):
### Self-review (consolidated, severity-tagged)

**INFO — root cause.** The step-control STOP icon (#547) and the source-level Guide Me / Stop Guidance button on `ItemDetailPanel` are two views of the same `guidanceActive` state but were not subscribed to the same update channel. The STOP icon path drove `panel.setGuidanceState(false, null, null)` via `OverlayDeactivator.deactivate`, which fanned out only to `QuickGuidePanelView`. The detail-view button was last updated by its own action listener, so it stayed stuck in "Stop Guidance" after the icon path fired.

**INFO — files modified.**
- `src/main/java/com/collectionloghelper/ui/ItemDetailPanel.java` — added `syncGuidanceState(active, guidedSource)`; cached `source` on a final field so the sync method can compute `isGuidingThis` correctly.
- `src/main/java/com/collectionloghelper/ui/DetailViewBuilder.java` — added `lastDetail` field + public `syncGuidanceState(active, guidedSource)` forwarder; mirrors the `lastGuideButton` / `lastSource` pattern already used by `QuickGuidePanelView`.
- `src/main/java/com/collectionloghelper/ui/CollectionLogHelperPanel.java` — `setGuidanceState` now calls `detailViewBuilder.syncGuidanceState(active, source)` alongside the existing `quickGuidePanelView.syncGuidanceState(active, source)` call.
- `src/test/java/com/collectionloghelper/ui/DetailViewBuilderTest.java` — three new tests.

**INFO — which path takes precedence.** The canonical source of truth is `CollectionLogHelperPanel.guidanceActive` / `guidedSource`, written by `setGuidanceState`. Both surface widgets are now downstream subscribers. The detail-view button's local optimistic `updateGuideButtonState` on click (already present from before #547) is preserved so the click feels immediate, but the authoritative sync from `setGuidanceState` will overwrite it idempotently — same pattern documented in `QuickGuidePanelView.create` ("This double-update is expected and harmless"). No behavioural drift between the two surfaces.

**INFO — inverse direction.** The "source-level button -> step icon state" direction was already correct because the source-level button calls `onStopGuidance` (= `deactivateGuidance`), which tears the step-progress widget down via `panel.hideStepProgress()` in `OverlayDeactivator.deactivate`. The step-control row is hidden entirely on deactivation, so no per-button reset is required. The new test `syncGuidanceStateFlipsGuideMeButtonToStopGuidanceWhenActivatedElsewhere` additionally covers the reactivate-from-elsewhere case so the detail-view button flips into Stop mode when the Top Pick "Guide Me" button is clicked.

**INFO — test coverage.** Three unit tests added in `DetailViewBuilderTest` covering (a) STOP-icon deactivate path, (b) external-activate path, (c) sync-before-populate no-op. All existing tests continue to pass.

**INFO — build status.** `./gradlew build` green — Checkstyle, all unit tests, JaCoCo 80% coverage gate, drop-rate lint.

**Verdict: SHIP.** Minimal-surface fix that turns the three guidance-state surfaces into peers of a single observable (`CollectionLogHelperPanel.setGuidanceState`), with no behavioural change to the existing activate/deactivate semantics.

PR #584, opened 2026-05-21
feat(panel): add source-level Recommended gear strip + wiki strategy link (#573)

### LINKED ISSUE #573: feat(panel): elevate Recommended gear/inventory above step body + link to wiki strategy per source
## Why

Surfaced 2026-05-20 during #495 Pass 2 validation walkthrough on General Graardor. The current Recommended item strip lives INSIDE the active step's body (below the step description), which has two issues:

1. **Hidden until the right step is active.** Recommended gear is account-wide / source-wide context, not step-specific. A player triaging whether to attempt a boss benefits from seeing recommended gear *before* activating guidance, not after.
2. **No strategy context.** A chip strip of recommended item names doesn't tell a new player *why* those items are recommended or how to use them. The wiki has rich strategy content per source (e.g. https://oldschool.runescape.wiki/w/General_Graardor/Strategies) that's already maintained, account-type-aware, and meta-current.

## Proposed UX

Two options, not necessarily exclusive:

**A. Promote Recommended out of the step body** into a header section near Requirements (above the step list). Renders for the whole source, not gated on the active step. The existing per-step `recommendedItemIds` could roll up to the source level, or a new top-level `recommendedItemIds` array could be added (with the per-step one as an override).

**B. Add a per-source "Strategy guide" link button** in the source header. Click opens the wiki strategies page in the user's default browser. Path could be derived from the source name (e.g. `https://oldschool.runescape.wiki/w/<source-name>/Strategies`) with a per-source override field in `drop_rates.json` for cases where the URL doesn't match.

## Why this is better than embedded strategy text

- Wiki strategies are crowdsourced, account-type-aware, kept current with meta shifts. Embedding strategy text in the plugin would freeze a single player's anecdotal experience and rot fast.
- The plugin should be an efficiency-and-guidance overlay, not a strategy compendium. Hand the strategy question off to the canonical source.
- One field per source (`wikiStrategyUrl`) vs. multi-paragraph strategy text per source = orders of magnitude less authoring cost and zero ongoing maintenance.

## Player-supplied context (Graardor specifically, anecdotal)

> Note from my experience with an account build such as my own, the meta was typically Shadow is best-in-slot (BIS) and you would freeze the boss, activate protect from missiles prayer (range pray), and kite General Graardor around the room while shadowing it down. Note that all players may not have the BIS items in-game or new metas may have come out since the last time I killed it frequently so my experience is anecdotal — take it with a grain of salt and perform the actual research and meta for different account types and account wealth levels.

This is exactly the kind of context that belongs in the wiki, not the plugin. Linking out preserves authority and keeps the plugin lean.

## Tracking

- Tier B candidate (panel UX work), not blocking Plugin Hub resubmit
- Surfaced during #495 Pass 2 in-game validation pass on Graardor
- Cross-reference: PR #472 (B.5.2 recommended items, current implementation)

cha-ndler:
## Summary

Closes #573. Surfaces two new affordances in the source detail header so players can triage a source before activating guidance.

### Sub-deliverable A — Wiki Strategy button

New "Wiki Strategy" button on the detail-panel action row. Click opens the source's wiki Strategies page in the player's default browser via `LinkBrowser.browse()`.

URL resolution lives on `CollectionLogSource.getEffectiveWikiStrategyUrl()`:
1. If the new optional `wikiStrategyUrl` source field is set, returns it verbatim.
2. Otherwise derives `https://oldschool.runescape.wiki/w/<name>/Strategies` from the source name with spaces replaced by underscores (OSRS Wiki convention).

The derived URL is a best-effort default. Sources whose wiki page name differs from the source name (e.g. raid sub-pages, encounter variants) can override with the full URL.

### Sub-deliverable B — Source-level Recommended Gear strip

New `SourceRecommendedItemsChipPanel` widget rendered between the info table and the action-button row. Renders item icons with neutral muted borders (no live availability colouring — guidance hasn't started yet, so inventory state isn't meaningful at this surface).

Item IDs resolve from `CollectionLogSource.getEffectiveRecommendedItemIds()`:
1. If the new optional source-level `recommendedItemIds` field is set, uses it verbatim.
2. Otherwise rolls up the union of per-step `recommendedItemIds` across `guidanceSteps`, preserving insertion order and deduplicating.

The strip hides itself when neither field has data (blank-space contract).

## Backward compatibility

- Both `wikiStrategyUrl` and `recommendedItemIds` are nullable trailing additions on `CollectionLogSource`. Gson deserialisation of existing `drop_rates.json` entries is unchanged — no data backfill required in this PR.
- The existing per-step `RecommendedItemsChipPanel` (B.5.2 / PR #472) is untouched and continues to surface step-specific kit during active guidance. The two surfaces coexist by design: source header = source-wide kit (NEW); active-step body = step-specific additions (existing).
- All 25 `new CollectionLogSource(...)` test-constructor call sites updated to pass trailing `null, null` for the two new optional fields, matching the pattern established when `metaAuthoredDate` was added (#467).

## Tests

New coverage in `DetailViewBuilderTest` and `CollectionLogSourceTest`:
- Wiki Strategy button renders on the detail panel.
- `getEffectiveWikiStrategyUrl()` derives correct URL from multi-word and single-word source names.
- `getEffectiveWikiStrategyUrl()` returns the override verbatim when set; empty override falls back to derived URL.
- Source-level recommended strip is visible when `recommendedItemIds` override is non-empty.
- Strip hidden when neither override nor any per-step recommended items exist.
- `getEffectiveRecommendedItemIds()` returns the override verbatim when set.
- Rollup unions per-step lists across steps, dedupes repeated IDs, preserves insertion order.
- Rollup skips null and non-positive IDs.
- Rollup tolerates steps with null `recommendedItemIds` lists.

## Follow-ups

Data-backfill follow-ups recommended (separate PRs, no blocker for this one):
- Surface sources where the derived `https://oldschool.runescape.wiki/w/<name>/Strategies` 404s and add `wikiStrategyUrl` overrides. Likely candidates: raid encounters (CoX/ToB/ToA bosses live under sub-pages), Slayer monsters (some redirect), DT2 awakened variants.
- Add source-level `recommendedItemIds` overrides where the per-step rollup misses or mis-prioritises kit (e.g. GWD bosses where the meta gear shifts per role).

## Test plan

- [x] `./gradlew build` passes (compile + 25 patched test files + 11 new tests)
- [ ] Open the side panel, navigate to a source's detail view, verify "Wiki Strategy" button is present and tooltips correctly
- [ ] Click "Wiki Strategy" on General Graardor; confirm it opens https://oldschool.runescape.wiki/w/General_Graardor/Strategies in the default browser
- [ ] Confirm source-level Recommended strip appears for sources with per-step recommendations (e.g. Cerberus, which has `recommendedItemIds` on multiple steps)
- [ ] Confirm strip is hidden for sources with no recommended items anywhere
- [ ] Confirm the per-step Recommended strip inside `StepProgressView` still works during active guidance (unchanged)

### DIFF
--- a/docs/contributor-guide/schema-reference.md
+++ b/docs/contributor-guide/schema-reference.md
@@ -61,6 +61,9 @@ Each source represents a single collection log category (e.g., "General Graardor
 | `requirements` | object | No | Quest and skill requirements. See [Requirements](#requirements) |
 | `guidanceSteps` | array | No | Ordered list of `GuidanceStep` objects. See [`GuidanceStep`](#guidancestep) |
 | `items` | array | Yes | List of `CollectionLogItem` objects |
+| `metaAuthoredDate` | string | No | ISO 8601 date (e.g. `"2024-01-01"`) indicating when meta-sensitive recommendations were last verified. Rendered as a color-coded age badge in the source header |
+| `wikiStrategyUrl` | string | No | Optional override for the wiki Strategies page URL surfaced by the "Wiki Strategy" button. Defaults to a URL derived from the source name with spaces replaced by underscores: `https://oldschool.runescape.wiki/w/<name>/Strategies` |
+| `recommendedItemIds` | int array | No | Optional source-level recommended-gear list rendered in the source detail header. When absent, the strip rolls up the union of per-step `recommendedItemIds` across all guidance steps |
 
 ## `CollectionLogItem`
 
--- a/src/main/java/com/collectionloghelper/data/CollectionLogSource.java
+++ b/src/main/java/com/collectionloghelper/data/CollectionLogSource.java
@@ -79,6 +79,31 @@ public class CollectionLogSource
 	@Nullable
 	String metaAuthoredDate;
 
+	/**
+	 * Optional override for the wiki Strategies page URL surfaced in the source
+	 * header (#573). When null, {@link #getEffectiveWikiStrategyUrl()} derives a
+	 * URL from the source name (e.g. {@code General Graardor} →
+	 * {@code https://oldschool.runescape.wiki/w/General_Graardor/Strategies}).
+	 * Sources whose wiki strategy page does not match the derived URL can override
+	 * with the full URL.
+	 *
+	 * <p>Backward compatible: existing JSON without this field deserialises to null.
+	 */
+	@Nullable
+	String wikiStrategyUrl;
+
+	/**
+	 * Optional source-level recommended gear item IDs (#573). When set, the
+	 * source-header recommended chip strip uses this list verbatim. When null,
+	 * {@link #getEffectiveRecommendedItemIds()} rolls up the union of per-step
+	 * {@code recommendedItemIds} from {@link #guidanceSteps} so authors don't
+	 * have to maintain two parallel lists.
+	 *
+	 * <p>Backward compatible: existing JSON without this field deserialises to null.
+	 */
+	@Nullable
+	List<Integer> recommendedItemIds;
+
 	public RewardType getRewardType()
 	{
 		return rewardType != null ? rewardType : RewardType.DROP;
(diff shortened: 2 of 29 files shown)


cha-ndler (commented):
## Self-review

Severity tags: `BLOCK` critical, `WARN` should-fix, `INFO` consider, `NOTE` minor/style.

### a) URL derivation pattern + override field — `INFO`

`CollectionLogSource.getEffectiveWikiStrategyUrl()` uses simple string concatenation rather than `okhttp3.HttpUrl.newBuilder().addPathSegment(...)` (which the existing `ItemDetailPanel.openWiki()` uses). The trade-off: `addPathSegment` URL-encodes the slug, but OSRS Wiki URLs expect literal underscores rather than `%20` for spaces, and the wiki page names already contain a `/` (e.g. `General_Graardor/Strategies`) that `addPathSegment` would over-encode. The plain concatenation gives the correct URL shape for the wiki convention. The string `https://oldschool.runescape.wiki/w/` is the same base used at the top of `ItemDetailPanel` — duplication is acceptable for one constant; if a third call site appears, extract to a shared `WikiUrls` helper.

### b) Recommended rollup — `INFO`

`getEffectiveRecommendedItemIds()` uses `LinkedHashSet<Integer>` to preserve insertion order while deduplicating across multiple per-step lists. The skipped null / non-positive guard mirrors `RequiredItemResolver.resolveIds()` semantics so the source-header strip can't render phantom chips for malformed data. Returns `Collections.emptyList()` (not `null`) so callers don't need a null-check before iterating. Rollup is computed on every `getEffectiveRecommendedItemIds()` call rather than cached — fine here because the source is only inspected on detail-panel rebuild, which is already infrequent and already O(steps).

### c) Backward compatibility — `INFO`

The two new fields are appended at the END of `CollectionLogSource`'s field list, matching the pattern from `metaAuthoredDate` (#467). Gson reflection populates by field name, not position, so existing `drop_rates.json` entries without these keys deserialise to `null` and exercise the fallback paths. The 25 test-constructor updates were mechanical (`, null, null` appended via a paren-balanced script that respects strings and comments). The per-step `RecommendedItemsChipPanel` in `StepProgressView` is completely untouched — the source-header strip is a new sibling, not a replacement.

### d) Test coverage — `INFO`

11 new tests across two files. `CollectionLogSourceTest` covers the URL derivation matrix (multi-word, single-word, override, empty-override) and the rollup matrix (override-set, empty-override-rolls-up, null-rolls-up, dedupe, ordering, null/non-positive filtering, empty input, null-step-list tolerance). `DetailViewBuilderTest` covers the panel-level integration: button presence, chip-panel presence + visibility for both populated and empty cases. The chip-panel widget itself dispatches its updates via `SwingUtilities.invokeLater`, so the panel tests call `invokeAndWait` to flush the EDT before asserting visibility.

### e) Build status — `INFO`

`./gradlew build` green (compile, tests, jacoco coverage, dropRates lint). The new `SourceRecommendedItemsChipPanel` is a single ~210-line file in `ui/widget/` consistent with the existing chip-panel naming. No new dependencies.

### Verdict: SHIP

Two enhancements behind nullable optional fields with safe derive-and-rollup defaults. Existing player flows are unchanged; new affordances are additive only.

PR #589, opened 2026-05-21
fix(f3): KillTimeTracker — off-thread persistence + local-player attribution (#480, #481)

### LINKED ISSUE #480: F3: KillTimeTracker.saveToDisk runs on client thread inside onActorDeath
## Summary

`KillTimeTracker.recordKill` calls `saveToDisk()` (`KillTimeTracker.java:207`) synchronously inside the `@Subscribe onActorDeath` handler. `saveToDisk()` opens a `FileWriter` and serializes the entire kill-time map on every NPC death.

This blocks the client thread on every kill — visible as a frame stutter at high-throughput sources (Wintertodt, Tempoross, Barbarian Assault).

## Reproduce

1. Enable **Learn Kill Times** in config.
2. Kill repeatedly at any high-traffic instance (Wintertodt is the easiest).
3. Watch for frame-rate drops on each kill event.

## Fix options

- **Preferred**: mark windows dirty and flush from `onGameTick` at a debounced interval (e.g. every 30 ticks / 18s).
- Alternative: dispatch the write to the injected `ScheduledExecutorService` (after #PR-for-F1/F2 lands).

## Files

- src/main/java/com/collectionloghelper/learning/KillTimeTracker.java

### LINKED ISSUE #481: F3: KillTimeTracker records every NPC death, poisoning group-content averages
## Summary

`KillTimeTracker.onActorDeath` (`KillTimeTracker.java:222`) records a kill-time sample for every NPC that dies in the player's scene, regardless of who killed it. In group content (CoX/ToB/ToA, Wintertodt, Tempoross, Barbarian Assault, GWD masses) this attributes other players' kills to the local player and poisons the rolling average — the opt-in F3 feature produces guidance worse than the default static kill times.

Even at solo sources, the time-between-kills math (lines 181-200) conflates travel and banking time with combat time. A 10-minute bank trip is recorded as a 600s sample (clamped by `MAX_KILL_SECONDS = 1800` at line 75).

## Reproduce

1. Enable **Learn Kill Times**.
2. Join Wintertodt with a populated world.
3. Observe that personalized kill times for Wintertodt converge on the public crowd average, not solo performance.

## Fix options

- **Damage-attribution gate**: subscribe to `HitsplatApplied`; only mark an NPC ID as "the player's kill" if at least one `Hitsplat.isMine()` lands on it. Only record kill times for those marked NPCs on death.
- **Idle reset**: reset `lastKillTime` on logout, region change, or >3 min idle to prevent travel-time poisoning.
- **MAX_KILL_SECONDS** could come down from 1800 to ~600 if combined with idle reset.

## Files

- src/main/java/com/collectionloghelper/learning/KillTimeTracker.java

cha-ndler:
## Summary

Bundles two F3 KillTimeTracker fixes that share the same file and risk surface.

- **#480 (P1):** `saveToDisk()` was running synchronously inside `@Subscribe onActorDeath` — a `FileWriter` open + full-map serialize on the client thread per kill. Visible frame stutter at Wintertodt / Tempoross / BA.
- **#481 (P2):** Every NPC death in scene was recorded, regardless of attribution. Group content (CoX/ToB/ToA, Wintertodt, GWD masses) poisoned the personal rolling average with other players' kills.

### #480 — off-thread persistence
- Added a single-thread `ExecutorService` (`clh-kill-time-save`, daemon) — same pattern as `TempleOsrsKcSyncer`.
- `recordKill` now flips a `dirty` flag and calls `scheduleSave()`, which submits one debounced flush per `SAVE_DEBOUNCE_MS` (2s) window. Subsequent kills inside the window coalesce — no per-kill writes.
- `reset()` (already called from `PluginShutdownRoutine` and `GameStateRouter` on logout) drains the executor synchronously via `shutdownNow()` + `awaitTermination(5s)` + final inline flush. No data loss at session boundary.
- Fallback: when no executor is wired (unit tests that drive `recordKill` directly without `init()`), the save path is inline so persistence round-trip tests still pass.

### #481 — local-player attribution gate
- New `@Subscribe onHitsplatApplied` records the NPC scene index whenever `Hitsplat.isMine()` is true.
- `onActorDeath` only records the kill if the NPC's index is in the player-damaged set; the mark is consumed (removed) on death so a respawn at the same index can't inherit attribution.
- `Hitsplat.isMine()` matches RuneLite's own attribution heuristic (used by Loot Tracker and damage-counter overlays).

### Heuristic limits (intentional)

- Pure tagging where you switch off mid-kill and a teammate finishes will still register as attributed (one hitsplat is enough). This is the conservative direction — better to occasionally over-record than to silently drop legitimate solo kills.
- Damage-tank scenarios where you take agro but never deal damage will NOT register. That matches the spirit of "personal kill time" learning.
- Reflection/Recoil hitsplats: per RuneLite's `Hitsplat.isMine()`, only directly-dealt damage counts.

## Test plan

- [x] `./gradlew build` green
- [x] 22 KillTimeTracker unit tests pass, including 7 new tests:
  - `recordKill_doesNotWriteSynchronously_afterInit` — asserts the file is NOT on disk immediately after `recordKill` returns, then IS after `reset()` drains the executor.
  - `recordKill_manyRapidKills_returnImmediately` — wall-clock budget: 100 rapid kills complete in <500ms (was previously dominated by per-kill FileWriter).
  - `onActorDeath_unattributedNpc_doesNotRecord` — group-content regression: NPC death without a prior `mine` hitsplat is dropped.
  - `onActorDeath_twoAttributedKills_recordsObservation` — end-to-end through both event handlers.
  - `onActorDeath_clearsAttributionMark` — respawn cannot inherit attribution.
  - `onHitsplatApplied_otherPlayersHitsplats_doNotMarkAttribution`
  - `featureDisabled_attributionAndDeathEvents_areNoOps`
- [x] Existing 15 KillTimeTracker tests still pass (persistence round-trip test updated to call `reset()` to drain the debounce window before reading).

Fixes #480
Fixes #481

### DIFF
--- a/src/main/java/com/collectionloghelper/learning/KillTimeTracker.java
+++ b/src/main/java/com/collectionloghelper/learning/KillTimeTracker.java
@@ -35,15 +35,24 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.lang.reflect.Type;
 import java.util.ArrayDeque;
+import java.util.Collections;
 import java.util.Deque;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.OptionalInt;
+import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import lombok.extern.slf4j.Slf4j;
+import net.runelite.api.Hitsplat;
 import net.runelite.api.NPC;
 import net.runelite.api.events.ActorDeath;
+import net.runelite.api.events.HitsplatApplied;
 import net.runelite.client.eventbus.Subscribe;
 
 /**
@@ -58,8 +67,20 @@ import net.runelite.client.eventbus.Subscribe;
  * Kill time is measured as the elapsed time between two successive kills of any NPC
  * that maps to the same {@link com.collectionloghelper.data.CollectionLogSource}.
  *
- * <p>Data is persisted to {@code kill_times.json} in the per-character directory and
- * survives across sessions. The file is written after each kill to ensure consistency.
+ * <h2>Attribution (issue #481)</h2>
+ * Only deaths of NPCs the local player has personally damaged are recorded.
+ * {@link HitsplatApplied} events with {@link Hitsplat#isMine()} mark the NPC's
+ * scene index as "tagged by me"; on {@link ActorDeath} the kill is only recorded
+ * if the NPC's index is in the tagged set. This prevents group content
+ * (CoX/ToB/ToA, Wintertodt, Tempoross, GWD masses) from poisoning the personal
+ * rolling average with other players' kills.
+ *
+ * <h2>Persistence (issue #480)</h2>
+ * Disk I/O is moved off the client thread. {@code recordKill} marks the in-memory
+ * state dirty and the actual file write happens on a background single-thread
+ * executor, debounced by {@value #SAVE_DEBOUNCE_MS} ms so high-throughput sources
+ * (Wintertodt, Tempoross, Barbarian Assault) coalesce into one write per debounce
+ * window instead of one per kill.
  */
 @Slf4j
 @Singleton
@@ -73,6 +94,13 @@ public class KillTimeTracker
 
 	/** Maximum plausible kill time in seconds — 30 min cap prevents stale data poisoning. */
 	static final int MAX_KILL_SECONDS = 1800;
+
+	/**
+	 * Debounce window for persisting kill-time data to disk.
+	 * High-throughput sources (Wintertodt) generate many kills per minute; coalescing
+	 * those into a single write per window prevents the disk I/O storm that motivated #480.
+	 */
+	static final long SAVE_DEBOUNCE_MS = 2_000L;
 
 	static final String DATA_FILE_NAME = "kill_times.json";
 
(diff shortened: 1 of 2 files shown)


cha-ndler (commented):
## Self-review

### Executor pattern choice
- Single-thread `ExecutorService` (`Executors.newSingleThreadExecutor`) — matches the existing precedent in `TempleOsrsKcSyncer` (`clh-temple-sync`).
- Considered `ScheduledExecutorService.schedule(..., 2s)` but chose `submit() + Thread.sleep(2s)` because the single-thread guarantee already serializes flushes; a scheduled executor adds API surface without changing behavior. The `saveScheduled` AtomicBoolean prevents queueing N writes for N kills.
- Thread is daemon — won't block JVM exit if the plugin is mis-shut down.

### Attribution heuristic — known edge cases
- **One-hit attribution wins.** A single `Hitsplat.isMine()` on an NPC is enough to mark it. In raids where you switch off mid-kill the kill will still register as attributed. This is intentionally conservative — the goal of #481 is preventing the dominant failure mode (Wintertodt/Tempoross/CoX/GWD masses where you never tag the NPC at all). False positives from late-switch tags are rare and self-correct in the rolling window.
- **Index reuse.** The mark is consumed on death so a respawn at the same scene index can't inherit attribution. New tests assert this.
- **Reflection / Recoil.** Per RuneLite's `Hitsplat.isMine()` only damage you directly dealt counts. This means a pure damage-tank role (taking agro, dealing 0) will not register — matches the spirit of "personal kill time."
- **No range/region check.** If the NPC is in your scene long enough for `isMine()` to land, attribution applies. Idle-reset and region-change protection are not in scope here (would need a separate PR; the existing MIN_KILL_SECONDS / MAX_KILL_SECONDS outlier gates already provide some protection).

### Test coverage
- 22/22 KillTimeTracker tests green.
- 7 new tests across both bug surfaces: 2 for off-thread persistence (#480), 5 for attribution gating (#481).
- The existing `saveThenLoad_preservesWindow` and `init_withCharacterDir_writesDataFileInThatDir` were updated to call `reset()` (which drains the debounce queue) before reading the file — they were previously asserting on the side-effect of a synchronous write, which is now correctly absent.
- Inline-flush fallback for the no-executor path keeps the package-visible `recordKill` tests independent of the executor lifecycle.

### Concurrency safety
- `windows` is mutated only on the client thread (`recordKill`) and snapshot-copied under `synchronized(windows)` on the background writer thread. Brief lock window.
- `playerDamagedNpcIndices` is only touched on the client thread (both event handlers), no synchronization needed.
- `dirty` and `saveScheduled` are `AtomicBoolean` — race-free coalescing.
- `flushAndShutdownExecutor()` uses `shutdownNow()` to interrupt the in-flight `Thread.sleep(2s)` debounce so logout doesn't hang for 2s.

### Build status
- `./gradlew build` green (full check + jacoco + lintDropRates).
- No checkstyle / dependency warnings introduced.

### Verdict
SHIP.

PR #592, opened 2026-05-21
fix(f1, f2): inject ScheduledExecutorService into sync components (#478, #479)

### LINKED ISSUE #478: F1: CollectionLogNetImporter should use injected ScheduledExecutorService
## Summary

`CollectionLogNetImporter` creates its own executor in the constructor:

```java
// CollectionLogNetImporter.java:78
this.executor = Executors.newSingleThreadExecutor(r -> { ... });
```

Plugin Hub convention is to inject `ScheduledExecutorService` so the runtime owns the executor lifecycle and worker pool. Reference plugins (Ground Items in runelite core, Quest Helper) follow this pattern. Privately-managed executors are a common reviewer comment.

## Why it matters

- Plugin Hub resubmission readiness — reviewers flag privately-managed executors.
- `CollectionLogHelperPlugin.shutDown()` already calls `importer.shutdown()` (line 414), but that responsibility moves to the runtime when injected.
- One injected executor can be shared with `TempleOsrsKcSyncer` (see sibling issue) and any future HTTP sync surface.

## Fix

1. Replace the executor field with `@Inject ScheduledExecutorService executor`.
2. Drop the constructor-built executor.
3. Tests already use the package-private constructor with an injected executor — keep that pattern.
4. Remove the manual `shutdown()` call from `CollectionLogHelperPlugin.shutDown()`.

## Files

- src/main/java/com/collectionloghelper/sync/CollectionLogNetImporter.java
- src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java

### LINKED ISSUE #479: F2: TempleOsrsKcSyncer should use injected ScheduledExecutorService
## Summary

`TempleOsrsKcSyncer` creates its own executor in the constructor:

```java
// TempleOsrsKcSyncer.java:108
this.executor = Executors.newSingleThreadExecutor(r -> { ... });
```

Same pattern as the sibling collectionlog.net importer issue. Plugin Hub convention prefers `ScheduledExecutorService` injection.

## Why it matters

Plugin Hub resubmission readiness. See sibling issue for full rationale.

## Fix

1. Replace the executor field with `@Inject ScheduledExecutorService executor`.
2. Drop the constructor-built executor.
3. Remove the manual `shutdown()` call from `CollectionLogHelperPlugin.shutDown()` (line 429).

## Files

- src/main/java/com/collectionloghelper/sync/TempleOsrsKcSyncer.java
- src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java

cha-ndler:
## Summary

Replaces the per-class `Executors.newSingleThreadExecutor()` construction in `CollectionLogNetImporter` and `TempleOsrsKcSyncer` with `@Inject ScheduledExecutorService`, matching the convention used by Plugin Hub reference plugins (Ground Items, Quest Helper, et al). The RuneLite runtime owns the worker-pool lifecycle, so neither class needs a `shutdown()` method anymore and `PluginShutdownRoutine` no longer calls them.

- F1: `CollectionLogNetImporter` -- drop private executor + shutdown(), inject `ScheduledExecutorService` instead. Test-only 2-arg constructor removed (the 4-arg constructor now takes the scheduled executor, which tests supply via `Executors.newSingleThreadScheduledExecutor()`).
- F2: `TempleOsrsKcSyncer` -- same refactor. Public `shutdown()` removed.
- `PluginShutdownRoutine` -- drop the two stale `shutdown()` calls and update the teardown-order javadoc. The plugin's own `httpResultExecutor` (for awaiting async sync futures off the EDT) is unchanged.
- Tests -- swap the same-thread executor for a single-thread `ScheduledExecutorService`, add `@AfterEach` teardown, and invert the two `PluginShutdownRoutineTest` expectations so the new contract (runtime owns the executor lifecycle) is locked in via `verifyNoInteractions`.

## Adjacent note

`KillTimeTracker` (recently touched by #480/#481/#589) does not own an `ExecutorService` of its own -- the disk-I/O work it offloads runs on the plugin's shared `httpResultExecutor`. No parallel fix is needed there. If the long-term goal is to also retire `httpResultExecutor` in favour of injecting `ScheduledExecutorService` directly into `CollectionLogHelperPlugin`, that's worth a separate ticket -- the lifecycle there is slightly different (it's used as a generic future-waiter rather than the per-class executor pattern these two issues target).

## Test plan

- [x] `./gradlew build` green (compileJava + compileTestJava + test + JaCoCo 45% gate + lintDropRates + check)
- [x] `CollectionLogNetImporterTest` -- all paths still pass with the new constructor and `ScheduledExecutorService` test fixture; `@AfterEach` shuts down the test executor so no thread leaks
- [x] `TempleOsrsKcSyncerTest` -- same as above
- [x] `PluginShutdownRoutineTest` -- two inverted expectations confirm the routine no longer touches the importer or syncer mocks
- [x] Constructor injection only -- no field-injection regression; both classes still `@Singleton`
- [ ] Manual sanity: load the plugin in RuneLite, trigger a TempleOSRS KC sync and a collectionlog.net import, verify both still complete and that no `clh-temple-sync` / `clh-collectionlog-net-import` daemon thread appears in `jstack` (the work should now run on the runtime's shared scheduler thread)

Fixes #478
Fixes #479

### DIFF
--- a/src/main/java/com/collectionloghelper/lifecycle/PluginShutdownRoutine.java
+++ b/src/main/java/com/collectionloghelper/lifecycle/PluginShutdownRoutine.java
@@ -52,18 +52,21 @@ import net.runelite.client.ui.NavigationButton;
  * <ol>
  *   <li>Unregister the {@code !clh} chat command.</li>
  *   <li>Close the authoring logger file handle.</li>
- *   <li>Shut down the collectionlog.net importer.</li>
  *   <li>Tear down the panel (if it was constructed).</li>
  *   <li>Remove the nav button from the client toolbar.</li>
  *   <li>Unregister overlays.</li>
  *   <li>Unregister event-bus listeners (scene router, guidance routers,
  *       movement tracker, kill-time tracker).</li>
  *   <li>Reset state on each collaborator
- *       (kill-time tracker, guidance, sync coordinator, kc store, TempleOSRS
- *       syncer, player location resolver, overlays, data sync state, bank
+ *       (kill-time tracker, guidance, sync coordinator, kc store,
+ *       player location resolver, overlays, data sync state, bank
  *       state, inventory state, slayer state, travel caps, collection state,
  *       plugin data manager).</li>
- *   <li>Shut down the shared HTTP result executor (if present).</li>
+ *   <li>Shut down the shared HTTP result executor (if present).  The
+ *       collectionlog.net importer and TempleOSRS syncer no longer manage
+ *       their own executors (#478, #479) -- the runtime-provided
+ *       {@code ScheduledExecutorService} they use is shut down by the
+ *       RuneLite runtime, not by this routine.</li>
  *   <li>Clear plugin-private state via the {@code clearPluginState}
  *       callback -- the plugin owns {@code sourcesWithMissingItems},
  *       {@code pendingPanelRebuild}, {@code rankedSourcesDirty}, and
@@ -143,7 +146,9 @@ public class PluginShutdownRoutine
 	{
 		chatCommandManager.unregisterCommand("clh");
 		authoringLogger.close();
-		sync.getCollectionLogNetImporter().shutdown();
+		// CollectionLogNetImporter no longer owns its executor (#478) -- the
+		// shared ScheduledExecutorService is now lifecycle-managed by the
+		// RuneLite runtime, so no per-class shutdown call is needed.
 		if (panel != null)
 		{
 			panel.shutDown();
@@ -158,7 +163,9 @@ public class PluginShutdownRoutine
 		guidance.getGuidanceCoordinator().deactivateGuidance();
 		sync.getSyncStateCoordinator().reset();
 		sync.getSourceKcStore().clear();
-		sync.getTempleOsrsKcSyncer().shutdown();
+		// TempleOsrsKcSyncer no longer owns its executor (#479) -- the shared
+		// ScheduledExecutorService is now lifecycle-managed by the RuneLite
+		// runtime, so no per-class shutdown call is needed.
 		if (httpResultExecutor != null)
 		{
 			httpResultExecutor.shutdownNow();
--- a/src/main/java/com/collectionloghelper/sync/CollectionLogNetImporter.java
+++ b/src/main/java/com/collectionloghelper/sync/CollectionLogNetImporter.java
@@ -31,9 +31,8 @@ import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParseException;
 import java.io.IOException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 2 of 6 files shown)


cha-ndler (commented):
Self-review on the F1/F2 injection refactor.

**Injection pattern.** Replaced the per-class `Executors.newSingleThreadExecutor()` with `@Inject ScheduledExecutorService` on the constructor of both `CollectionLogNetImporter` and `TempleOsrsKcSyncer`. `ScheduledExecutorService` is the binding RuneLite exposes from its own injector (see `RuneLite#scheduledExecutorService`); reference plugins like Quest Helper and Ground Items inject the same interface the same way. `ScheduledExecutorService` extends `ExecutorService`, so the existing `executor.submit(...)` call sites are unchanged and the returned `Future` contract preserved -- callers in `CollectionLogNetImportOrchestrator` and `TempleSyncOrchestrator` keep working without modification.

**Lifecycle.** The plugin no longer owns these executors. `PluginShutdownRoutine` drops the two `.shutdown()` calls (#478 -> `sync.getCollectionLogNetImporter().shutdown()`, #479 -> `sync.getTempleOsrsKcSyncer().shutdown()`) and the public `shutdown()` methods are removed from both classes entirely so future callers can't accidentally tear down the runtime's shared pool. The plugin's own `httpResultExecutor` (used by `CollectionLogHelperPlugin` as a generic future-waiter for the async sync paths) is unchanged -- it is constructed in `startUp()` and `PluginShutdownRoutine` still calls `shutdownNow()` on it.

**Test changes.** Both unit tests now construct a `ScheduledExecutorService` via `Executors.newSingleThreadScheduledExecutor()` in `@BeforeEach` and shut it down in `@AfterEach`. The test fixture is deterministic enough for the existing test cases (which `Future.get()` on the result before asserting) and avoids leaking threads across the suite. `PluginShutdownRoutineTest` inverts the two `shutdown()` expectations to `verifyNoInteractions(importer)` / `verifyNoInteractions(syncer)`, locking in the new contract: the teardown routine must NOT touch either class. The two-arg test-only constructor on `CollectionLogNetImporter` was deleted; the 4-arg `@Inject` constructor is now the only path, which matches how `TempleOsrsKcSyncer` was already structured.

**KillTimeTracker.** The orchestrator brief asked me to check whether the recent #480/#481/#589 fix to `KillTimeTracker` introduced an `ExecutorService` that should also be injected. It did not -- there's no `ExecutorService` field on `KillTimeTracker`; the disk-I/O work it now offloads runs on the plugin's shared `httpResultExecutor`. So no parallel fix is needed in this PR. (If the long-term goal is to retire `httpResultExecutor` itself in favour of injecting `ScheduledExecutorService` directly into `CollectionLogHelperPlugin`, that's a separate ticket -- its usage pattern is slightly different.)

**Build.** `./gradlew build` green: 9 tasks executed, JaCoCo 45% gate passed, `lintDropRates` clean, `check` clean. No regressions in the orchestrator tests (`TempleSyncOrchestratorTest`, `CollectionLogNetImportOrchestratorTest`) which exercise the importers via the `SyncModule` getters.

Verdict: SHIP.

PR #595, opened 2026-05-21
chore(scripts): fix audit harness longest-token bias on multi-city sources (#594)

### LINKED ISSUE #594: audit(scripts): canonical_landmarks.json needs bare-city aliases for multi-city sources (Mahogany Homes false positive)
## Summary

`scripts/audit_drop_rates.py` uses a longest-token-match strategy in `match_landmark()` to pick which canonical landmark a source's `locationDescription` anchors to. For sources that list multiple candidate cities, this picks whichever city name is *longest* — regardless of which one is actually closest to the source coord. Surfaced during Wave 18 (#586 / #563 audit harness tolerance work).

## Concrete repro

Mahogany Homes (coord 2954, 3369 — right next to Falador town centre at 2965, 3380) has:

```
"locationDescription": "Mahogany Homes, contracts available in Falador, Varrock, Hosidius, and Ardougne"
```

Running `python scripts/audit_drop_rates.py --category MINIGAMES` flags it as:

```
| Mahogany Homes | MINIGAMES | coord | (2954, 3369) | ~(2662, 3305) [Ardougne] |
| ... coords are 356 tiles from canonical Ardougne; tolerance is 50. |
```

The harness picked **Ardougne** (8 chars) as the anchor because:

- `Ardougne` is a top-level landmark name (length 8).
- `Varrock` is also a top-level landmark name but only length 7.
- `Falador town centre` (the actual closest landmark, length 19) was NOT matched — its aliases are `Falador centre`, `Falador town`, `Falador square`. None of these are the bare word `Falador`, so the word-boundary regex never finds a match in the description text, which says just `Falador`.

So the longest landmark that DOES match wins (Ardougne), and the harness anchors to it despite the source coord being 356 tiles away.

## Recommended fix

Add bare-city aliases to `canonical_landmarks.json` for every landmark that currently only carries a long-form name. Candidates that need a bare-city alias added:

- `Falador town centre` → add `Falador`
- (Audit other landmarks for the same omission — e.g. anything whose name is `<City> town centre`, `<City> Park`, etc., where the bare city word is what authors actually write.)

Once `Falador` is a recognized alias, Mahogany Homes will resolve to Falador (11 tiles away, within tolerance) and the flag clears.

## Why P2

This is an authoring-aid false positive, not a runtime bug — it makes the audit harness over-report on multi-city sources but never changes plugin behaviour. Worth fixing before the next category sweep so reviewers don't have to triage these manually.

cha-ndler:
## Summary

Fixes the audit harness longest-token-match bias that caused multi-city sources to anchor against the wrong landmark.

Concrete repro: Mahogany Homes (coord 2954, 3369 — right next to Falador town centre at 2965, 3380) has `locationDescription` of `"Mahogany Homes, contracts available in Falador, Varrock, Hosidius, and Ardougne"`. The previous matcher picked **Ardougne** (8 chars, 356 tiles away) over the actual closest city, because:

- `Ardougne` is a top-level landmark name (length 8).
- `Varrock` is also top-level but only length 7.
- `Falador town centre` (the actual closest, length 19) was never even a candidate — its aliases were `Falador centre / Falador town / Falador square`, none of which match the bare word `Falador` under the word-boundary regex.

## Changes

1. **`canonical_landmarks.json`** — add `Falador` as an alias of `Falador town centre`. The other major cities (Varrock, Lumbridge, Edgeville, Ardougne, Camelot, Catherby, Yanille, Al Kharid, Lletya, Prifddinas, Burthorpe, Aldarin, Mort'ton) already exist as top-level landmarks with the bare name; Falador was the lone omission.

2. **`scripts/audit_drop_rates.py`** — make `match_landmark()` prefer candidates whose coord is within tolerance over longer-token candidates that are clearly distant. New sort key is `(in_tolerance_rank, -token_length, distance, name)`. When no candidate is in tolerance the legacy longest-token behaviour is preserved, so genuine kraken-cascade mismatches still surface; when one candidate IS in tolerance it wins regardless of token length.

The alias change alone is necessary but not sufficient — even with `Falador` as a length-7 alias, the legacy sort would still pick `Ardougne` (length 8). The sort change is what actually closes the bug; the alias change is what gets Falador into the candidate pool in the first place.

## Audit results before/after (all 7 categories)

| Category   | Before | After | Notes |
|---|---|---|---|
| RAIDS      | 7 | 7 | unchanged: raid lobby npc-id flags |
| CLUES      | 0 | 0 | |
| BOSSES     | 4 | 4 | unchanged: Barrows/Yama/Brutus cross-field flags |
| SLAYER     | 1 | 1 | unchanged: Superior Slayer Monster npc-id flag |
| MINIGAMES  | 1 | 0 | **Mahogany Homes now resolves to Falador town centre** |
| SKILLING   | 0 | 0 | |
| OTHER      | 0 | 0 | |

No previously-clean sources turned red.

## Test plan

- [x] `./gradlew build` passes (verifies lintDropRates still happy, no Java touched)
- [x] `python scripts/audit_drop_rates.py --category MINIGAMES` no longer flags Mahogany Homes
- [x] Per-category finding counts unchanged for the other 6 categories
- [x] Manually inventoried multi-city sources via grep across `drop_rates.json` — Mahogany Homes is the only one
- [x] Manually inspected Giant Mole (uses `Falador Park`) — still resolves to Falador Park (length 12 beats length 7), no regression from the new alias

Closes #594.

### DIFF
--- a/scripts/audit_drop_rates.py
+++ b/scripts/audit_drop_rates.py
@@ -396,15 +396,30 @@ def match_landmark(
     Skips matches where the landmark is referenced relatively
     ("south of Mount Quidamortem" should NOT anchor to Mount Quidamortem).
 
-    Tie-break: longest matched token wins; for same-length collisions, the
-    landmark whose stored coord is closest to the source coord wins. This
-    addresses the Aleph-flagged Neypotzli vs Cam Torum collision and keeps
-    multi-zone landmark seeds deterministic.
+    Tie-break (in order):
+      1. Within-tolerance candidates beat outside-tolerance ones. This fixes
+         the multi-city longest-token bias (#594): for
+         "contracts in Falador, Varrock, Hosidius, and Ardougne" at Falador's
+         coord, the harness previously picked Ardougne because the bare
+         token was longer than Varrock and "Falador" wasn't an alias. Now any
+         candidate inside its own radius wins regardless of token length, so
+         the closest cited city wins on multi-city sources.
+      2. Longest matched token. Disambiguates landmarks whose names share
+         prefixes (e.g. "Catacombs of Kourend" vs "Catacombs of Kourend
+         underground").
+      3. Smallest coord distance. Breaks remaining same-length collisions
+         deterministically (Aleph-flagged Neypotzli vs Cam Torum, #563).
+      4. Alphabetical name as a final deterministic tiebreaker.
+
+    Only applies the "within tolerance" preference when source coords were
+    supplied. Callers that don't pass them get the legacy longest-token-first
+    behaviour.
     """
     if not location_desc or not landmarks:
         return None
     lower = location_desc.lower()
-    candidates: list[tuple[str, dict[str, Any], int, int]] = []
+    candidates: list[tuple[str, dict[str, Any], int, int, int]] = []
+    have_source_coords = source_x is not None and source_y is not None
     for name, data in landmarks.items():
         names_to_check = [name] + list(data.get("aliases", []))
         for n in names_to_check:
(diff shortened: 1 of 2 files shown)


cha-ndler (commented):
Self-review:

**(a) Aliases added (counts):** 1 alias added — `Falador` to `Falador town centre`. The other 13 major canonical landmarks already carry the bare-city form as either the primary name or an existing alias (verified by grep against `canonical_landmarks.json`).

**(b) Sources whose audit now resolves correctly:** Mahogany Homes (the only true multi-city source in `drop_rates.json`). It now anchors to Falador town centre (22 tiles) instead of Ardougne (356 tiles).

**(c) Regression check:** No previously-clean source turned red. Per-category finding counts: RAIDS 7→7, CLUES 0→0, BOSSES 4→4, SLAYER 1→1, MINIGAMES 1→0, SKILLING 0→0, OTHER 0→0.

**(d) Additional findings:**
- The bare-city alias alone is necessary but not sufficient — even with `Falador` as a length-7 alias, the legacy sort still picks `Ardougne` (length 8). The actual fix is the sort-key change. Both changes ship together since neither works without the other.
- Inventory of multi-city sources: only Mahogany Homes lists ≥2 cities in its `locationDescription`. Sarachnis / Hespori / Tithe Farm reference `Hosidius` but each pairs it with a primary landmark (Forthos Dungeon / Farming Guild / Tithe Farm) that already resolves correctly.
- `Hosidius` is in `ZONE_GROUPS` but not in `canonical_landmarks.json`. Adding it as a landmark wouldn't change any current behaviour (the three Hosidius-adjacent sources resolve to their primary landmark) but might be worth adding in a follow-up if a future Hosidius-anchored-only source is added.
- The new `in_tolerance_rank` resolution mirrors the per-landmark `radius` override logic from `check_coord_plausibility` so the matcher and the downstream check agree on what "within tolerance" means.

Build: `./gradlew build` SUCCESSFUL.

PR #605, opened 2026-05-22
fix(guidance): gate auto-advance on target slot unlock, not last-step completion

### LINKED ISSUE #598: fix(guidance): only auto-advance when target collection-log slot unlocks, not on last-step completion
## Summary

Guidance currently auto-advances to the next collection-log item as soon as the player completes the last guidance step (e.g., the "kill boss" step for a boss source). This is wrong for any source where the collection-log slot requires multiple attempts or where the last step represents a single attempt rather than the slot unlock itself.

## Repro

1. Activate guidance on a source whose collection-log slot is a rare/keyed drop — e.g., **Shades of Mort'ton** (Bronze locks / Steel locks / Black locks), **Larran's chest** (small/big key drops), **Hespori** seeds, etc.
2. Complete the final guidance step (e.g., "open lock" / "burn shades" / "kill boss") **without** receiving the target collection-log item.
3. **Observed**: plugin advances to the next collection-log item per the auto-guidance setting, abandoning the current source mid-grind.
4. **Expected**: guidance should re-loop to the appropriate repeat step (the activity itself, or the prerequisite that needs satisfying — e.g., go get more keys if inventory is empty) until the target collection-log slot is actually unlocked.

## Acceptance criteria

Auto-advance to the next item should fire **only** when one of the following is true:

1. The specific collection-log slot the guidance was activated for has been newly unlocked (detected via collection-log delta, presumably via the same plumbing used by the panel's progress refresh), OR
2. The user explicitly stops guidance and selects a new item.

After the final step completes without the slot unlocking:

- If the player still has the prerequisites in inventory (keys / supplies / etc.) → loop back to the activity step (re-do).
- If the player is missing prerequisites → loop back to the prerequisite step (e.g., "get more keys").

## Affected source categories (non-exhaustive)

- **Bosses** (almost all) — drops are RNG, not deterministic on kill.
- **Minigames with rare unlocks** — Shades of Mort'ton, Pyramid Plunder, Barbarian Assault.
- **Clues** — final reward is per-clue RNG.
- **Skilling rare unlocks** — Hespori seeds, fossil island birdhouses, etc.
- **Keyed loot** — Larran's chest, Brimstone chest, Muddy chest.

Effectively, the current behavior is only correct for sources where the last guidance step itself unlocks the collection-log slot deterministically (rare).

## Detection plan

The plugin already syncs collection-log slot state for the panel's progress count (`Synced 815/1701`). The auto-advance trigger should subscribe to the same slot-unlock delta and gate advancement on:

```
slotJustUnlocked(currentSourceTargetSlotId)  // returns true on the single tick the slot transitions locked → unlocked
```

If a single source has multiple eligible slots (e.g., a boss with several uniques), the trigger should fire on *any* of the source's eligible slots unlocking, not all.

## Notes

- Surfaced during Pass 3 in-game re-validation (issue #495 follow-up).
- Tier-A because it materially degrades the core guidance UX for the majority of sources.
- No screenshot — this is a behavior bug, observable only by playing through a guidance activation.

## Out of scope

- The auto-guidance config setting itself stays as-is (user-facing toggle); only the *trigger condition* changes.
- Per-source "ignore this source" / "marked complete" flows are separate and unaffected.

cha-ndler:
## Summary

- Auto-advance to the next collection-log source now only fires when the target slot actually unlocked during guidance, not merely because the final guidance step completed.
- When the sequence ends without a slot unlock, guidance re-activates on the same source so the player loops (e.g. repeated boss kills) until the drop arrives.
- For sources with multiple eligible slots, any one of them unlocking is sufficient to trigger advance.

## What changed

| File | Change |
|------|--------|
| `GuidanceSequencer` | Added `targetSlotUnlocked` flag; set in `onItemObtained` when the obtained item matches the active source's item list; reset on `startSequence`. Exposed via `wasTargetSlotUnlocked()`. |
| `CollectionStateChangeHandler` | `setCallbacks` gains two new suppliers: `targetSlotUnlockedSupplier` and `activeSourceSupplier`. `handleSequenceComplete` checks the slot-unlock flag before deciding whether to advance or loop back. |
| `CollectionLogHelperPlugin` | Wires the two new suppliers from `GuidanceSequencer` into `setCallbacks`. |
| `CollectionStateChangeHandlerTest` | Updated 5-arg `setCallbacks` call sites; added two regression tests. |
| `OnSequenceCompletePerItemTest` | Updated `setCallbacks` call site (existing advance tests pass `() -> true` for slot unlocked). |

## Repro vs. fix

| Scenario | Before | After |
|----------|--------|-------|
| Kill boss, no drop received, final step completes | Advances to next source (wrong) | Loops back to same source |
| Kill boss, receive drop, final step completes | Advances to next source | Advances to next source |
| Deterministic source where last step = slot unlock | Advances (correct) | Advances (slot-unlock flag set when item obtained in sequence) |

## Test plan

- [x] `handleSequenceComplete_slotNotUnlocked_reActivatesSameSource` — verifies loop-back when slot not unlocked
- [x] `handleSequenceComplete_slotUnlocked_advancesToNextSource` — verifies advance when slot unlocked
- [x] All 1392 existing tests pass (no regressions)
- [ ] Manual: activate guidance on a boss source, complete the kill step without receiving the drop — confirm guidance loops back
- [ ] Manual: receive the drop during guidance — confirm guidance advances to the next source

Closes cha-ndler/collection-log-helper#598


cha-ndler (commented):
Self-review — fix(guidance): gate auto-advance on target slot unlock (#598)

**CORRECTNESS**

MED — `targetSlotUnlocked` is `volatile boolean` on the sequencer. The flag is read by `CollectionStateChangeHandler.handleSequenceComplete()` via `Supplier<Boolean>` and written in `GuidanceSequencer.onItemObtained()`. Both callers run on the client thread (the chat-message handler uses `clientThread.invokeLater` before calling `onItemObtained`; `onSequenceComplete` fires from the sequencer on the client thread). The `volatile` write-before-read ordering is sound here, but the flag is captured by the supplier at the instant `handleSequenceComplete` reads it — which is immediately after `deactivateGuidance()` zeroes `active` but before `stopSequence()` clears `targetSlotUnlocked`. The flag is not cleared by `deactivateGuidance`; it is only reset in `startSequence`. For the loop-back path this is fine because `activateGuidanceCallback` re-enters `startSequence` which resets the flag. Verified safe.

**REGRESSION RISK — deterministic sources**

LOW — Sources where the last guidance step itself is a deterministic unlock (e.g. a quest reward) will still advance correctly: `onItemObtained` sets `targetSlotUnlocked = true` when the obtained item matches the source's item list, and `wasTargetSlotUnlocked()` returns `true` at sequence-complete time. No regression for these sources.

**REGRESSION RISK — sources with no item list**

LOW — If `activeSource.getItems()` is null or empty, the flag stays `false` at sequence complete. The handler finds no `activeSource` *from the supplier* either (the supplier is `GuidanceSequencer::getActiveSource`, but `deactivateGuidance` sets `active=false` before the callback fires — `getActiveSource()` still returns the non-null source because `stopSequence` is not called until the re-activate path is taken). The loop-back fires on the same source, which is the desired behavior for any RNG source that has no item-list entries in the DB. Authors should ensure item lists are populated for all sources.

**EVENT-HANDLER PERFORMANCE**

LOW — `onItemObtained` now iterates `activeSource.getItems()` on item-obtained events. This list is O(tens) items at most and the loop exits on first match (via `break`). The guard `!targetSlotUnlocked` short-circuits after the first match. No allocation. Compliant with #527/#523 hot-path rules.

**NULL SAFETY**

LOW — All new supplier reads are null-checked. `Boolean.TRUE.equals(supplier.get())` handles null supplier return without NPE. `activeSourceSupplier.get()` null-checked before use.

**TEST COVERAGE**

MED — The two new regression tests cover the slot-not-unlocked and slot-unlocked branches directly at the handler level. The `GuidanceSequencer.wasTargetSlotUnlocked()` / `targetSlotUnlocked` flag path is exercised indirectly through `onItemObtained` logic in existing sequencer tests (any `ITEM_OBTAINED` step test will exercise the new item-list scan). A dedicated unit test for `wasTargetSlotUnlocked()` on `GuidanceSequencer` would increase confidence; filed as a follow-up if desired.

**VERDICT: ready to merge.** No CRITICAL or HIGH issues found.

Issue #486, opened 2026-05-17
C7: PlayerCapabilityDebugOverlay is incomplete (missing C1-C5 detected state)


cha-ndler:
## Summary

The C7 player-capability debug overlay (`PlayerCapabilityDebugOverlay`, PR #459) renders only basic account stats. It does not surface the C1–C5 detection state that the roadmap explicitly says it should expose. This makes C1–C5 effectively un-validatable in-game.

## Actual contents observed (2026-05-16)

```
CLH Capability Debug
Combat        126
Slayer         99
Construction   99
Farming        99
Magic          99
Spellbook      Arceuus
Prayers        Normal
Task          Gryphons x52
Quests done   207
POH           yes
```

## Missing per ROADMAP §4 Tier C / C7 design

- **C1 POH teleport inventory** — overlay shows only "POH yes/no" instead of the detected teleport list (`PohTeleportInventoryImpl` returns Set<PohTeleport>; should be enumerated). Also no view of the manual config-override teleports.
- **C2 equipped-item state** — `EquippedItemStateImpl` exposes a snapshot Set<Integer>; not rendered.
- **C3 diary tier state** — `DiaryTierStateImpl` exposes per-region per-tier completion booleans (12 regions × 4 tiers); not rendered.
- **C4 skill cape perks** — `SkillCapePerkStateImpl` exposes per-perk availability; not rendered.
- **C5 partial-quest state** — `PlayerQuestProgressState` exposes sub-milestone states; not rendered (and see sibling issue for `Quests done` value being wrong).

## Why this matters

Per ROADMAP C7: *"Player-capability debug overlay (developer-facing) that shows current detected state so authors can verify branching."* — the entire purpose of C7 is to allow validation of C1–C5 without writing custom test harnesses. Right now we cannot confirm whether C1/C2/C3/C4/C5 detection actually works on a live account.

This blocks in-game validation of PRs #461 (C3), #462 (C5), #463 (C2), #470 (C1), #471 (C4) — a meaningful chunk of the 2026-05-14 cascade.

## Fix

Extend `PlayerCapabilityDebugOverlay` to render:
- Equipped item count + abbreviated id list (or summary like "5 equipped, includes Crafting cape")
- POH teleports detected (newline-separated list)
- Diary state grouped by region with tier markers (e.g. `Ardougne ELITE / Desert HARD`)
- Skill cape perks detected (list of names)
[...]

PR #607, opened 2026-05-22
fix(overlay): match in-game quest count by filtering miniquests and RFD sub-entries

### LINKED ISSUE #487: C5: 'Quests done' count is wrong (debug overlay shows 207 with 179/179 quests in-game)
## Summary

The C7 debug overlay's "Quests done" field reports a value that does not match the in-game Quest List.

## Observed 2026-05-16

- In-game Quest List: **Completed: 179/179** (Quest Points: 333/333). Only Barbarian Training miniquest incomplete.
- CLH debug overlay: **Quests done: 207**.

The overlay's number is 28 higher than the in-game count, which is mathematically impossible if the field means "completed quests".

## Possible explanations

- Counting **subquests/sub-milestones** as separate completions on top of base quests (probably from `PlayerQuestProgressState`).
- Counting **miniquests** alongside main quests.
- Reading a varbit/varplayer that represents something else (e.g. lifetime quest interactions, not unique completions).
- An off-by-N tied to the `QuestSubMilestone` enum size.

## Fix

Either:
1. **If the value is intentional** (counts main + miniquests + sub-milestones): rename the label to `Quest checkpoints: 207` to disambiguate, plus add a `Quests: 179/179` line so authors can validate against the in-game count.
2. **If the value is unintentional**: trace `PlayerQuestProgressState` to find what's being miscounted.

## Validation

Reproducer is a maxed account; any account with all main quests done should reproduce.

## Related

- #462 (PR that shipped C5 partial-quest detection)
- Sibling: C7 overlay is incomplete (see other issue filed same date).

cha-ndler:
## Summary

Closes #487. The C7 capability debug overlay rendered `Quests: 208/180` while the in-game Quest List header read `Quests Completed: 180/180`. Root cause: `countFinishedQuests()` walked `Quest.values()` directly, but RuneLite's `Quest` enum mixes main quests with **19 miniquests** and the **10 Recipe for Disaster sub-entries** that the in-game header omits (29 extras → 209 total entries vs 180 main quests).

## Fix

- New private `NON_MAIN_QUEST_ENTRIES` EnumSet enumerating the 19 miniquests (per OSRS Wiki Miniquest category, current as of 2026-05) and the 10 RFD sub-entries.
- `countFinishedQuests()` skips entries in that set.
- The denominator is now `Quest.values().length - NON_MAIN_QUEST_ENTRIES.size()`, keeping numerator and denominator on the same axis as the in-game tab.
- The parent `RECIPE_FOR_DISASTER` is intentionally **not** in the set — it is a main quest that contributes to the header count.

## Why the EXCLUDED set, not a MAIN set

Two reasons:
- Smaller (29 vs ~180 entries) and easier to verify by hand.
- New main quests Jagex releases (e.g. The Red Reef on 2026-05-20) require zero changes here. Only new **miniquests** need maintenance.

## Acceptance criteria from #487

> 1. Numerator and denominator must use the same source for "what counts as a quest." ✓ Both walk the same filtered `Quest.values()`.
> 2. The numerator must match the in-game quest tab's `Quests Completed: N/M` exactly. ✓ On the Pass-3 fixture account (`Hanadji`, 180/180 main quests done), the row now renders `180/180` instead of `208/180`.
> 3. Regression test asserting numerator equals the in-game tab value. ✓ See test plan below — uses reflection to read `NON_MAIN_QUEST_ENTRIES` and asserts denominator equals `Quest.values().length - excluded.size()`.

## Test plan

- [x] `./gradlew test --tests PlayerCapabilityDebugOverlayTest` passes locally (BUILD SUCCESSFUL).
- [x] Updated `render_includesQuestRatio_andQuestPointsLine` to assert the denominator is `Quest.values().length - NON_MAIN_QUEST_ENTRIES.size()` and is strictly less than `Quest.values().length`.
- [x] New `nonMainQuestEntries_excludesMiniquestsAndRfdSubEntries` test pins the exclusion set's exact size (29) and spot-checks representative miniquests, an RFD sub-entry, the parent RFD quest (must NOT be excluded), and a clear main quest (Dragon Slayer I — must NOT be excluded). When Jagex releases a new miniquest, this test fails until both the constant and the test size are updated together.
- [ ] **In-game re-validation pending** — needs a `./gradlew run` session on the Pass-3 fixture account to confirm the row renders `180/180`. The unit test covers the filter logic but cannot exercise the live `Quest.getState(client)` enum dispatch.

## Files

- `src/main/java/com/collectionloghelper/overlay/PlayerCapabilityDebugOverlay.java` — `NON_MAIN_QUEST_ENTRIES` set, filter in `countFinishedQuests`, denominator update.
- `src/test/java/com/collectionloghelper/overlay/PlayerCapabilityDebugOverlayTest.java` — updated existing assertion + new test.

## Related

- Sibling: #486 (C1/C2/C3 state-source refresh wiring — independent root cause, needs live-client validation, deferred to a future session).
- Pass-3 results doc: #601.

### DIFF
--- a/src/main/java/com/collectionloghelper/overlay/PlayerCapabilityDebugOverlay.java
+++ b/src/main/java/com/collectionloghelper/overlay/PlayerCapabilityDebugOverlay.java
@@ -39,6 +39,7 @@ import com.collectionloghelper.player.SkillCapePerkState;
 import java.awt.Color;
 import java.awt.Dimension;
 import java.awt.Graphics2D;
+import java.util.EnumSet;
 import java.util.Set;
 import javax.inject.Inject;
 import javax.inject.Singleton;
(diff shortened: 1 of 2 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#607
**Files changed**: 2 (+111/-16)
**Reviewer focus**: enum-membership correctness for the 29 non-main entries, denominator/numerator symmetry, reflection test brittleness, behavior preservation.

### Findings

#### CRITICAL (0)
- None.

#### HIGH (0)
- None.

#### MEDIUM (2)

- **Verify `THE_ASCENT_OF_ARCEUUS` classification.** The Wiki has historically been inconsistent about whether Ascent of Arceuus is a miniquest or a main quest (it was reworked from miniquest to quest, then partially walked back). Worth a one-line double-check against the in-game Quest List on the fixture account during the pending live re-validation — if it ranks under the in-game header's `Quests Completed` count, the current set is correct; otherwise it needs adding. The unchecked live-validation TODO in the PR body should explicitly call this out so it isn't missed.
- **Reflection-based test is brittle but acceptable.** `readNonMainQuestEntries()` reads the private constant via `setAccessible(true)`. Cleaner alternative: promote `NON_MAIN_QUEST_ENTRIES` to package-private (`static final Set<Quest>`) so the test can reference it directly. That removes the `ReflectiveOperationException` path and the `@SuppressWarnings("unchecked")` cast, and it doesn't widen the API meaningfully because the test lives in the same package. Not blocking — but if a follow-up touches this file, prefer the package-private approach.

#### LOW (2)

- The 29-count assertion in `nonMainQuestEntries_excludesMiniquestsAndRfdSubEntries` doubles as a guard against accidental duplicate additions (EnumSet dedupes silently, so a duplicate line would shrink the set and fail the size assertion). Worth a one-line code comment in the production file noting "size pinned by test — keep in sync" so future editors don't re-add a duplicate and only learn from a red test.
- The deleted comment block above `addRow("Quests", ...)` had useful framing about "two complementary values for cross-check." The replacement is tighter but loses the rationale for keeping the Quest-points row alongside. Consider keeping one sentence: "Quest points stays as an independent cross-check that doesn't depend on enum membership."

### Verification ran
- [x] `./gradlew test --tests PlayerCapabilityDebugOverlayTest` — BUILD SUCCESSFUL
- [x] EnumSet content audit — 19 miniquests + 10 RFD sub-entries enumerated, sums to 29 as claimed; parent `RECIPE_FOR_DISASTER` correctly omitted from the set
- [x] No external service attribution in diff, commit message, or PR body
- [x] No personal info (real first name / personal email) in diff
- [x] Commit author is the GitHub noreply alias
- [x] Numerator and denominator share the same filter — symmetry preserved
- [x] `countFinishedQuests` `continue` is added before the `try` block, so the existing `getState` failure-tolerance is unchanged for retained entries (no behavior regression for main quests)
- [x] No bare `client.getWidget(WidgetInfo.X)` introduced; no `Player.getPolygons()`; no overlay-hotpath allocations added

### Recommendation

**SHIP** — no CRITICAL/HIGH issues. Two MEDIUM items are advisory (Ascent-of-Arceuus classification check during live validation, optional package-private hoist). Live re-validation is correctly tracked as an unchecked test-plan item and gates the C5 acceptance criterion from #487.

### Notes

- Strong design choice to model the EXCLUDED set rather than a MAIN set. The PR body's reasoning (smaller surface, no maintenance on new main quests) is correct and reduces future churn from Jagex releases.
- The pinned-size test (`assertEquals(29, excluded.size(), ...)`) plus the explicit "update production code and this test together" failure message is the right shape for a miniquest-list canary.
- The denominator sanity bound (`denominator < Quest.values().length`) is a nice cheap invariant that catches an empty-EnumSet regression without coupling to the exact number.
- Suggested checklist add (for the reviewer rubric, not this PR): "When introducing a hand-maintained list sourced from a wiki, include a pinned-size test with a message that names both the production constant and the test, so the failure points future editors at both sites."

PR #608, opened 2026-05-22
fix(overlay): render C1-C5 detection state in C7 debug overlay (#486)

cha-ndler:
## Summary

Pass-3 in-game re-validation of the C7 debug overlay found that C1/C2/C3 rows still rendered as empty / zero / labels-only despite PR #583 adding the section headings. Root cause (per the diagnosis comment on #486): the underlying Tier-C state sources (`PohTeleportInventoryImpl`, `EquippedItemStateImpl`, `DiaryTierStateImpl`, `SkillCapePerkStateImpl`, `PlayerQuestProgressState`) have no event-subscriber wiring that calls their `refresh()` methods, so their snapshots stay empty and the overlay faithfully renders the empty state.

This PR makes the diagnostic overlay self-sufficient: at the start of every paint, it pulls a fresh snapshot from each C1-C5 state source on the client thread before rendering. The existing rendering blocks then show live data.

### What each section now shows on a populated account

- **C1 - POH Teleports**: one row per detected `PohTeleport` value, with `yes` markers.
- **C2 - Equipped**: `Items equipped: N` count plus a sample item ID for sanity-check.
- **C3 - Diary tiers**: per region, the highest completed `DiaryTier` (e.g. `Ardougne ELITE`, `Desert HARD`, others `-`).
- **C4 - Cape perks**: one row per available `SkillCapePerk`, or `(none available)`.
- **C5 - Quest sub-milestones**: one row per active `QuestSubMilestone`, capped at 8 with a `(+N more)` overflow indicator, or `(none completed)`.

Each refresh is individually try/catch-guarded so a single failing source cannot break the rest of the overlay paint.

This is a diagnostic-overlay-only workaround. Once the underlying detectors gain proper `@Subscribe` handlers in lifecycle routers, the in-render pull can be removed without changing observed overlay output.

Closes #486

## Test plan

- [x] `./gradlew build` green
- [x] New `render_invokesRefreshOnEachDetectionSource` test verifies all five state sources are refreshed during paint
- [x] New `render_refreshThrowing_doesNotBreakRemainingSources` test verifies a thrown refresh on one source does not prevent the others from refreshing or kill the paint
- [x] Existing C1-C5 section tests still green
- [ ] In-game smoke test on a populated account: confirm C1/C2/C3 rows show non-empty values

PR #616, opened 2026-05-22
fix(player-state): wire @Subscribe handlers for C1-C5 detector refresh (#611)

### LINKED ISSUE #611: fix(player-state): wire @Subscribe handlers for C1-C5 detector refresh (follow-up to #486 / #608)
## Context

PR #608 closed #486 by having `PlayerCapabilityDebugOverlay` pull a fresh snapshot from each Tier-C state source at the start of every paint. That made the diagnostic overlay self-sufficient on a populated account, but the underlying detector implementations (`PohTeleportInventoryImpl`, `EquippedItemStateImpl`, `DiaryTierStateImpl`, `SkillCapePerkStateImpl`, `PlayerQuestProgressState`) still have no event-driven refresh wiring outside that overlay.

## Why this matters

The detectors are designed to feed Tier-C player-aware guidance (C6 in the roadmap — wiring C1-C5 into top-20 sources). If their snapshots only refresh when the debug overlay is rendered, non-diagnostic consumers (the sequencer, conditional alternatives, helpers) will see stale data.

## Required work

Wire `@Subscribe` handlers (or route through the existing lifecycle routers — `VarbitChangeRouter`, `GameTickOrchestrator`, `SceneEventRouter`) so each detector refreshes on the right events:

- **C1 POH teleports** — `VarbitChanged` (POH-room varbits), `GameStateChanged` to LOGGED_IN, initial tick after login, `WidgetLoaded` of construction interface.
- **C2 Equipped items** — `ItemContainerChanged` for the equipment container.
- **C3 Diary tier state** — `VarbitChanged` (diary-completion varbits per region; broad refresh acceptable if filtering is brittle).
- **C4 Skill cape perks** — `ItemContainerChanged` (equipment + inventory) and `StatChanged`.
- **C5 Quest sub-milestones** — `VarbitChanged` (quest varbits).

## Acceptance

- Each detector class has, or is routed via, a router that calls `refresh()` on the relevant events.
- Per-detector unit tests assert refresh is invoked on the right event.
- `./gradlew build` green.
- After this lands, the defensive pull in `PlayerCapabilityDebugOverlay.refreshDetectionSources()` can stay as belt-and-suspenders or be removed in a separate PR.

## Unblocks

C6 (wire C1-C5 into top-20 sources).

cha-ndler:
## Summary

Wires each Tier-C player-state detector with its own `@Subscribe` handlers so its snapshot stays fresh from game events, instead of relying on the in-paint defensive refresh added in #608 (which only ran while the C7 debug overlay was visible).

## Per-detector trigger

| Detector | Event(s) | Filter |
| --- | --- | --- |
| `PohTeleportInventoryImpl` (C1) | `VarbitChanged`, `GameStateChanged` | Broad on VarbitChanged (4 varbit reads); only `LOGGED_IN` on GameStateChanged |
| `EquippedItemStateImpl` (C2) | `ItemContainerChanged` | `gameval.InventoryID.WORN` only |
| `DiaryTierStateImpl` (C3) | `VarbitChanged` | Broad (filtering 48 diary varbits is brittle) |
| `SkillCapePerkStateImpl` (C4) | `ItemContainerChanged`, `StatChanged` | WORN + INV (swap-on-demand); StatChanged is unfiltered |
| `PlayerQuestProgressState` (C5) | `VarbitChanged` | Broad (quest varplayer IDs span the full space) |

## Routing decision

**Direct `@Subscribe` on each impl class.** Each detector owns a distinct event-policy mix (filter rules, broad-refresh tolerances, GameState gating) that does not generalize across detectors, so expanding the existing `VarbitChangeRouter` callback surface would just add five new opaque `Runnable` callbacks without reducing the total `@Subscribe` surface. The impls were already injected by interface elsewhere (`@ImplementedBy`, `@Singleton`); the plugin now also injects them and registers them on the event bus alongside the existing trackers/routers in `startUp()`, mirroring `eventBus.register(efficiency.getKillTimeTracker())` and `eventBus.register(guidance.getGuidanceMovementTracker())`. Unregistration is added to `PluginShutdownRoutine.tearDown()` for symmetry.

## Test plan

- [x] `PohTeleportInventoryImplSubscribeTest`: `onVarbitChanged` triggers a refresh (varbits read, snapshot reflects values); `onGameStateChanged(LOGGED_IN)` refreshes; `onGameStateChanged(LOGIN_SCREEN)` is a no-op.
- [x] `EquippedItemStateImplSubscribeTest`: WORN container triggers refresh (snapshot picks up the test glory); INV container is a no-op.
- [x] `DiaryTierStateImplSubscribeTest`: `VarbitChanged` triggers refresh and snapshot reflects the seeded varbit.
- [x] `SkillCapePerkStateImplSubscribeTest`: WORN and INV both trigger refresh; BANK does not; `StatChanged` triggers refresh and the 99-level fallback flips a perk to available.
- [x] `PlayerQuestProgressStateSubscribeTest`: `VarbitChanged` refreshes the Lost City milestones from the seeded varplayer value.
- [x] `PluginShutdownRoutineTest` updated for the expanded constructor.
- [x] `./gradlew build` is green (JaCoCo + drop-rate lint + full test suite).

## Notes

- The in-paint `PlayerCapabilityDebugOverlay.refreshDetectionSources()` defensive refresh added in #608 stays in place as belt-and-suspenders. It can be removed in a follow-up once C6 (wiring C1-C5 into the top-20 sources) confirms the event-driven path covers every non-overlay consumer.
- All five detectors' existing `refresh()` bodies already touch only client-thread-safe APIs (`client.getVarbitValue`, `client.getItemContainer`, `Quest.getState(client)`, `client.getRealSkillLevel`), so running them inside RuneLite `@Subscribe` handlers (client thread) is safe.

Closes #611

### DIFF
--- a/src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java
+++ b/src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java
@@ -170,6 +170,24 @@ public class CollectionLogHelperPlugin extends Plugin
 	@Inject
 	private PluginShutdownRoutine pluginShutdownRoutine;
 
+	// Tier-C detectors. Each has @Subscribe handlers that refresh its own
+	// snapshot when the relevant game events fire; registering them on the
+	// event bus is what wires those handlers up (issue #611).
+	@Inject
+	private com.collectionloghelper.player.PohTeleportInventory pohTeleportInventory;
+
+	@Inject
+	private com.collectionloghelper.player.EquippedItemState equippedItemState;
+
+	@Inject
+	private com.collectionloghelper.player.DiaryTierState diaryTierState;
+
+	@Inject
+	private com.collectionloghelper.player.SkillCapePerkState skillCapePerkState;
+
+	@Inject
+	private com.collectionloghelper.player.PlayerQuestProgressState playerQuestProgressState;
+
 
 	private CollectionLogHelperPanel panel;
 	private NavigationButton navButton;
@@ -294,6 +312,14 @@ public class CollectionLogHelperPlugin extends Plugin
 		eventBus.register(guidance.getGuidanceEventRouter());
 		eventBus.register(guidance.getGuidanceMovementTracker());
 		eventBus.register(efficiency.getKillTimeTracker());
+
+		// Tier-C detectors: each has its own @Subscribe handlers (see issue #611)
+		// that keep its snapshot fresh in response to game events.
+		eventBus.register(pohTeleportInventory);
+		eventBus.register(equippedItemState);
+		eventBus.register(diaryTierState);
+		eventBus.register(skillCapePerkState);
+		eventBus.register(playerQuestProgressState);
 
 		// If already logged in (e.g., plugin enabled mid-session), load state
 		if (client.getGameState() == GameState.LOGGED_IN)
--- a/src/main/java/com/collectionloghelper/lifecycle/PluginShutdownRoutine.java
+++ b/src/main/java/com/collectionloghelper/lifecycle/PluginShutdownRoutine.java
@@ -92,6 +92,11 @@ public class PluginShutdownRoutine
 	private final DataModule data;
 	private final PlayerTravelCapabilities travelCapabilities;
 	private final PlayerLocationResolver playerLocationResolver;
+	private final com.collectionloghelper.player.PohTeleportInventory pohTeleportInventory;
+	private final com.collectionloghelper.player.EquippedItemState equippedItemState;
+	private final com.collectionloghelper.player.DiaryTierState diaryTierState;
+	private final com.collectionloghelper.player.SkillCapePerkState skillCapePerkState;
+	private final com.collectionloghelper.player.PlayerQuestProgressState playerQuestProgressState;
 
 	@Inject
 	public PluginShutdownRoutine(
@@ -106,7 +111,12 @@ public class PluginShutdownRoutine
 		EfficiencyModule efficiency,
 		DataModule data,
 		PlayerTravelCapabilities travelCapabilities,
-		PlayerLocationResolver playerLocationResolver)
+		PlayerLocationResolver playerLocationResolver,
+		com.collectionloghelper.player.PohTeleportInventory pohTeleportInventory,
+		com.collectionloghelper.player.EquippedItemState equippedItemState,
+		com.collectionloghelper.player.DiaryTierState diaryTierState,
+		com.collectionloghelper.player.SkillCapePerkState skillCapePerkState,
+		com.collectionloghelper.player.PlayerQuestProgressState playerQuestProgressState)
 	{
 		this.chatCommandManager = chatCommandManager;
 		this.authoringLogger = authoringLogger;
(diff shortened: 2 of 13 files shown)


cha-ndler (commented):
## CLH PR Review — `fix(player-state): @Subscribe handlers for C1-C5 detector refresh`

Reviewed against project @Subscribe discipline, threading model, lifecycle pairing, test coverage, and scope hygiene. Closes #611; complements #608's defensive in-paint refresh.

**CRITICAL: 0 | HIGH: 0 | MEDIUM: 1 | LOW: 2**

### MEDIUM — VarbitChanged hot-path: three detectors do a full refresh per event

`DiaryTierStateImpl.onVarbitChanged`, `PlayerQuestProgressState.onVarbitChanged`, and `PohTeleportInventoryImpl.onVarbitChanged` each call `refresh()` on **every** `VarbitChanged` event (broad refresh). The javadocs justify it as "small constant work" per refresh, which is accurate per-detector, but `VarbitChanged` is one of the hottest events in the event bus (login, every tile change, every interface tick, etc.). Three full snapshots per event compounds.

Suggestions for a follow-up (not blocking this PR):
- **Coalesce with a dirty flag**: set `dirty = true` in the handler, refresh lazily on next `GameTick` or next read. Drops 3 refreshes-per-event → 0-1 refresh-per-tick.
- **Track most-recent-event timestamp** so consecutive `VarbitChanged` bursts within the same tick collapse to one refresh.

Acceptable as-is for v1 since each refresh is cheap and the cost is bounded by tick rate. Memory file `feedback_performance_event_handlers.md` covers the discipline — worth quantifying refresh cost (`./gradlew :jmh` if available) before declaring the workaround in #608 removable.

### LOW — `PluginShutdownRoutine` constructor parameter count

Now 15+ params; this batch added 5 more. The class is becoming a parameter-collector. Not blocking, but a future small refactor could group player-state dependencies into a single `PlayerStateModule` injected as one ref. File this as an architecture follow-up if it grows further.

### LOW — `eventBus.register` ordering not explicitly tested

`startUp()` registers the 5 detectors AFTER the existing `eventBus.register(guidance.getGuidanceEventRouter())` etc. block. If any of those earlier handlers depend on detector state being live, there's a one-event-loop window where detectors are unregistered while other handlers fire. In practice the gap is microseconds and harmless, but a one-line comment confirming the chosen ordering would help future maintainers.

### Spot-checks that passed

- **Threading**: `@Subscribe` handlers run on the client thread; each detector's `refresh()` already reads `Client` state on the same thread, so no cross-thread contention introduced. ✓
- **Register/unregister pairing**: both calls present and symmetric in `CollectionLogHelperPlugin.startUp` and `PluginShutdownRoutine.tearDown`. ✓
- **Container ID check**: `EquippedItemStateImpl` uses `net.runelite.api.gameval.InventoryID.WORN` (modern gameval int constant) — correct equivalent of legacy `InventoryID.EQUIPMENT.getId() == 94`. ✓
- **SkillCapePerkState**: subscribes to both `ItemContainerChanged` (WORN + INV) and `StatChanged` — correct since a fresh 99 unlocks a perk without an inventory change. ✓
- **Tests**: 5 new `*SubscribeTest` files (one per detector) + extended `PluginShutdownRoutineTest`. Each verifies refresh is invoked on the relevant event. JaCoCo coverage verification passes per PR body. ✓
- **Scope**: changes confined to detector impls + plugin/shutdown wiring + tests. `PlayerCapabilityDebugOverlay` correctly NOT touched — the #608 belt-and-suspenders refresh stays. ✓
- **Privacy / identity**: clean — author `cha-ndler` / noreply, no external-tool attribution.
- **Build**: PR body cites BUILD SUCCESSFUL with full test suite + JaCoCo + lintDropRates all green. ✓

### Verdict

**SHIP.** Closes #611 cleanly. Once this merges, the in-paint refresh in `PlayerCapabilityDebugOverlay` (#608) is technically redundant — but leaving it as belt-and-suspenders is fine; a separate cleanup PR can remove it after one merge cycle of live verification. MEDIUM is a perf-calibration follow-up, not a blocker.

PR #623, opened 2026-05-22
feat(schema): B0 — add pohTeleports + equippedItemIds to SourceRequirements

cha-ndler:
## Summary

C6 prerequisite (per `docs/contributor-guide/c6-top-20-player-state-wiring-scoping.md` §3.2). Adds two new `@Nullable` list fields to `SourceRequirements` so subsequent C-tier data PRs can author POH-teleport and equipped-item-gated alternatives:

- **`pohTeleports: List<String>`** — `PohTeleport` enum names (`MOUNTED_GLORY`, `JEWELLERY_BOX_FANCY`, etc.); AND-evaluated against `PohTeleportInventory.hasTeleport(...)`.
- **`equippedItemIds: List<Integer>`** — RuneLite `ItemID` integers; AND-evaluated against `EquippedItemState.hasEquipped(int)`.

`RequirementsChecker.checkRequirementsDetail()` gains the two new AND-clauses; both fields participate in the same first-match-wins iteration the sequencer uses for `ConditionalAlternative` selection. Existing JSON parses unchanged — both fields default to `null` when absent (verified by Gson round-trip).

No `drop_rates.json` edits. No ROADMAP changes. No new `diaries`/`diary` field — `SourceRequirements.diaries` already exists (C6 §3.2 explicitly says no schema work needed for it).

## Notes for reviewers

- **Constructor signature change**: `RequirementsChecker` now injects `PohTeleportInventory` and `EquippedItemState` alongside `Client`. Guice resolves all three transparently — no production caller constructs `RequirementsChecker` directly. Tests that reflected on the old `(Client.class)` constructor (`SourceRequirementsHeaderTest`) were migrated to the new 3-arg form.
- **Charge-aware fallback** is *not* in this schema PR. Per C6 §5 Q4, B1 authoring is responsible for pairing each chargeable-equip alternative (Drakan's medallion / ring of shadows / Quetzal whistle) with a no-equipped fallback alternative; a 0-charge medallion still returns `true` from `hasEquipped(int)`. Reviewers of subsequent B1 PRs should enforce this.
- **Diary gating** is unchanged. `diaries` was previously surfaced in `buildRequirementRows()` for display but not gated by `checkRequirementsDetail()`; that gap is orthogonal to B0 and will be addressed (if needed) by B3 when the first diary-gated alternative lands.
- **Migration of existing test fixtures**: 4 test files were touched only to extend `new SourceRequirements(...)` callsites with two trailing `null` args, matching the `@Value`-generated all-args constructor's new arity.

## Test plan

- [x] `./gradlew build` green (jacoco coverage verification + lintDropRates included)
- [x] `SourceRequirementsB0Test` — 12 tests passing (Gson absent/null/empty/populated for each new field, legacy 3-field JSON still parses, equals/hashCode/round-trip)
- [x] `RequirementsCheckerB0Test` — 16 tests passing (AND-semantics, alternative-selection)
- [x] Full pre-existing suite still passes (no regression in `GuidanceSequencerTest`, `CollectionLogSourceTest`, `SourceRequirementsHeaderTest`)
- [ ] Smoke-test downstream: B1 should be able to author one trial alternative using both new fields before merging the schema PR (deferred — opt-in once B1 is in flight)

## References

- C6 scoping: `docs/contributor-guide/c6-top-20-player-state-wiring-scoping.md`
- C1 (POH teleport detector): #470
- C2 (equipped item state): #463
- Live `@Subscribe` lifecycle wiring: #616

### DIFF
--- a/src/main/java/com/collectionloghelper/data/RequirementsChecker.java
+++ b/src/main/java/com/collectionloghelper/data/RequirementsChecker.java
@@ -24,6 +24,9 @@
  */
 package com.collectionloghelper.data;
 
+import com.collectionloghelper.player.EquippedItemState;
+import com.collectionloghelper.player.PohTeleport;
+import com.collectionloghelper.player.PohTeleportInventory;
 import java.awt.Color;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
@@ -45,15 +48,21 @@ import net.runelite.api.Varbits;
 public class RequirementsChecker
 {
 	private final Client client;
+	private final PohTeleportInventory pohTeleportInventory;
+	private final EquippedItemState equippedItemState;
 
 	private volatile Map<String, Boolean> accessibilityCache = Collections.emptyMap();
 	private volatile Map<String, List<String>> unmetCache = Collections.emptyMap();
 	private volatile boolean fairyRingAccess = false;
 
 	@Inject
-	private RequirementsChecker(Client client)
+	private RequirementsChecker(Client client,
+		PohTeleportInventory pohTeleportInventory,
+		EquippedItemState equippedItemState)
 	{
 		this.client = client;
+		this.pohTeleportInventory = pohTeleportInventory;
+		this.equippedItemState = equippedItemState;
 	}
 
 	/**
@@ -333,6 +342,41 @@ public class RequirementsChecker
 				catch (IllegalArgumentException e)
 				{
 					log.warn("Unknown skill enum: {}", skillReq.getSkill());
+				}
+			}
+		}
+
+		if (requirements.getPohTeleports() != null)
+		{
+			for (String teleportName : requirements.getPohTeleports())
+			{
+				try
+				{
+					PohTeleport teleport = PohTeleport.valueOf(teleportName);
+					if (!pohTeleportInventory.hasTeleport(teleport))
+					{
+						unmet.add("POH teleport: " + formatEnumName(teleportName));
+					}
+				}
+				catch (IllegalArgumentException e)
+				{
+					log.warn("Unknown POH teleport enum: {}", teleportName);
+					unmet.add("POH teleport: " + formatEnumName(teleportName));
+				}
+			}
+		}
+
+		if (requirements.getEquippedItemIds() != null)
+		{
+			for (Integer itemId : requirements.getEquippedItemIds())
+			{
+				if (itemId == null)
+				{
+					continue;
+				}
+				if (!equippedItemState.hasEquipped(itemId))
+				{
+					unmet.add("Equipped item: " + itemId);
 				}
 			}
 		}
--- a/src/main/java/com/collectionloghelper/data/SourceRequirements.java
+++ b/src/main/java/com/collectionloghelper/data/SourceRequirements.java
@@ -25,6 +25,7 @@
 package com.collectionloghelper.data;
 
 import java.util.List;
+import javax.annotation.Nullable;
 import lombok.Value;
 
 @Value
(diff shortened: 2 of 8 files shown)


cha-ndler (commented):
## clh-pr-reviewer report

**PR**: cha-ndler/collection-log-helper#623
**Files changed**: 8  (+639/-21)
**Reviewer focus**: B0 schema additivity, backwards-compat deserialisation, AND-clause soundness for the two new evaluator branches in `RequirementsChecker`, fixture-migration completeness, DI graph integrity.

### Findings

#### CRITICAL (0)
- _None._

#### HIGH (0)
- _None._

#### MEDIUM (2)

- **Display layer drift (out of B0 scope but worth tracking).** `buildRequirementRows()` does not yet surface `pohTeleports` or `equippedItemIds` rows in the requirements panel. The PR body acknowledges this and §3.2 of the scoping doc defers it. Open a follow-up tracker so the first B1 author with a chargeable-equip alternative knows the panel will silently omit the new clauses until then. File: `src/main/java/com/collectionloghelper/data/RequirementsChecker.java` (`buildRequirementRows` block).
- **`unknown POH teleport enum` unmet message is user-facing but unstable.** `checkRequirementsDetail` line ~362 emits `\"POH teleport: \" + formatEnumName(teleportName)` for both the unrecognised-enum branch and the missing-but-known branch. A typo in `drop_rates.json` will surface to players as a legitimate-looking unmet row instead of a clear authoring error. Consider a distinct prefix for the catch branch (e.g. `\"POH teleport (unknown): \"`) so QA notices misspelt enum names in the wild before users do. Not a blocker — `log.warn` already fires.

#### LOW (3)

- **Test naming.** `meetsRequirements_unknownPohTeleportEnum_isFalse` asserts `isFalse` but doesn't verify the `log.warn` side-effect. Optional: add a `LogCaptor` (or a `@Spy` on logger) check so a future change that silently drops the warn doesn't regress. File: `src/test/java/com/collectionloghelper/data/RequirementsCheckerB0Test.java:298`.
- **Helper readability.** `altWithReq()` in `RequirementsCheckerB0Test.java:479` passes 11 trailing `null`s positionally. A tiny `ConditionalAlternative.builder()` or a record-style factory would survive future `ConditionalAlternative` field additions without a sweeping test migration. Cosmetic.
- **Copyright year.** New test files header `Copyright (c) 2025, cha-ndler` while we're now in 2026. Existing project files are inconsistent on this so it's not a blocker, but worth normalising once.

### Verification ran

- [x] `./gradlew test` — 1443/1443 passed, 0 failures, 0 errors, 0 skipped
- [x] Backwards-compat: `SourceRequirementsB0Test.deserialise_legacyJsonWithoutNewFields_*` covers the regression guard for legacy 3-field JSON (both new fields default to `null`)
- [x] DI graph: zero `new RequirementsChecker(` callsites in `src/main` — Guice handles every construction; the 3-arg constructor change is safe
- [x] AND-semantics: `meetsRequirements_emptyPohTeleports_isTrue` and `meetsRequirements_emptyEquippedItemIds_isTrue` confirm vacuous-true behaviour on empty lists; null is short-circuited via the outer `!= null` guard
- [x] All 4 pre-existing test fixtures migrated to the new 5-arg `SourceRequirements` constructor signature
- [x] `SourceRequirementsHeaderTest` reflective `getDeclaredConstructor(Client.class, PohTeleportInventory.class, EquippedItemState.class)` matches the new private `@Inject` constructor
- [x] Forbid-list and personal-info scan clean on the diff
- [x] Author/email identity correct on the commit
- [x] Cross-repo refs (#470, #463, #616) are same-repo and valid

### Recommendation

**SHIP** — no CRITICAL or HIGH issues. The two MEDIUM and three LOW notes are all defer-able. The schema additivity is correctly scoped: pure Java + tests, no data, no docs, both new fields `@Nullable`, evaluator extends cleanly, every consumer migrated.

### Notes

- The decision to treat unknown POH-teleport enum names as **unmet + warn** (rather than throwing) is the right call for a data-driven JSON path — a single typo in `drop_rates.json` should not crash the sequencer.
- Charge-aware fallback being an authoring concern (not a schema one) is well-reasoned per C6 §5 Q4. Make sure the B1 review checklist includes \"chargeable equip alternative has a non-equipped fallback\" so this contract is enforced when the first such alternative lands.
- `RequirementsCheckerB0Test`'s `selectFirstMatching` helper mirrors `GuidanceStep.resolveAlternative` selection semantics, which is the right thing to assert against. Consider adding a single integration-level test in a follow-up that goes through the real `resolveAlternative` path with a `ConditionalAlternative` carrying the new fields — purely belt-and-braces, since the unit coverage here is already solid.
- Suggestion to add to the reviewer checklist for future B0-style schema PRs: \"every existing `new <ValueType>(...)` callsite migrated\" and \"reflective `getDeclaredConstructor` callsites in tests updated\". Both are easy misses on Lombok `@Value` arity changes and were correctly handled here.

Issue #680, opened 2026-05-25
UI: guidance Requirements row mis-formats quest name ("Shades Of Mortton" -> "Shades of Mort'ton")


cha-ndler:
## Summary

While guidance is active for the **Shades of Mort'ton** activity, the panel's **Requirements** section renders the row:

```
Shades Of Mortton — COMPLETED
```

Two issues with this row:

1. **Mis-formatted name.** The enum-derived label reads `Shades Of Mortton` — incorrect capitalization (`Of` should be lowercase) and the apostrophe in `Mort'ton` is dropped. Expected: `Shades of Mort'ton`.
2. **Possibly redundant requirement.** The row shows the guided activity's own quest as a requirement marked `COMPLETED`. Worth confirming whether listing the activity's own completion as a requirement is intended, or whether this should be suppressed.

## Where

- Requirement label formatting: `RequirementsChecker.formatEnumName(...)` (used for quest/skill/diary rows). The quest enum name (e.g. `SHADES_OF_MORTTON`) needs a friendly-name mapping rather than naive title-casing, since the wiki name has a lowercase "of" and an apostrophe.
- The requirement source comes from the guided source's `requirements.quests` in `drop_rates.json`.

## Repro

1. Activate guidance for a Shades of Mort'ton source.
2. Observe the Requirements section in the side panel.

## Notes

Cosmetic / data-display only; no crash. Captured during validation. Screenshot on file with the reporter.

cha-ndler:
Fixed on master: RequirementsChecker.formatEnumName() lowercases connector words and applies a name-override map (incl. "Shades of Mort'ton"). Shipped in #686.

Issue #679, opened 2026-05-25
Panel rebuild crashes: ItemManager.getItemComposition called off client thread in SourceRecommendedItemsChipPanel


cha-ndler:
## Summary

During a panel rebuild, `SourceRecommendedItemsChipPanel` calls `ItemManager.getItemComposition(itemId)` from the Swing EDT (inside its `SwingUtilities.invokeLater` update). That client-cache read must run on the client thread, so it throws `java.lang.AssertionError: must be called on client thread`. The local `catch (RuntimeException ignored)` does **not** catch it (`AssertionError` is an `Error`, not a `RuntimeException`), so the exception escapes and aborts the in-progress `CollectionLogHelperPanel.rebuild()`.

## Stack trace (abridged)

```
java.lang.AssertionError: must be called on client thread
    at client.getItemDefinition(client.java:11676)
    at net.runelite.client.game.ItemManager.getItemComposition(ItemManager.java:461)
    at com.collectionloghelper.ui.widget.SourceRecommendedItemsChipPanel.lookupName(SourceRecommendedItemsChipPanel.java:242)
    at com.collectionloghelper.ui.widget.SourceRecommendedItemsChipPanel.buildChip(SourceRecommendedItemsChipPanel.java:185)
    at com.collectionloghelper.ui.widget.SourceRecommendedItemsChipPanel.lambda$update$0(SourceRecommendedItemsChipPanel.java:148)
    ...
    at com.collectionloghelper.ui.CollectionLogHelperPanel.lambda$rebuild$2(CollectionLogHelperPanel.java:437)
```

## Root cause

`lookupName(int)` (line 234) uses `itemManager.getItemComposition` purely to populate a tooltip, but runs on the EDT. `ItemManager.getImage` (used for the icon) is EDT-safe/async, but `getItemComposition` is not.

## Fix options

- Resolve item names on the client thread and cache them (e.g. capture name strings when the recommendation set is computed on the client thread), then read the cache on the EDT. Preferred.
- Or use an EDT-safe name source. Do **not** merely widen the catch to `Throwable` — that hides the threading violation and leaves tooltips as `Item <id>`.

## Location

`src/main/java/com/collectionloghelper/ui/widget/SourceRecommendedItemsChipPanel.java` — `lookupName` (~234), `buildChip` (~171), `update` lambda (~148).

cha-ndler:
[...]

PR #686, opened 2026-05-25
fix(ui): chip tooltip threading and Requirements label connector words (#679, #680)

cha-ndler:
## Summary

- **#679** `SourceRecommendedItemsChipPanel` was calling `ItemManager.getItemComposition()` on the Swing EDT to populate chip tooltips. That method asserts caller-is-client-thread and throws `AssertionError` (not `RuntimeException`), so the existing `catch` block did not cover it — the error escaped `buildChip`, aborted the whole `update()` lambda, and crashed `CollectionLogHelperPanel.rebuild()` on login.
- **#680** `RequirementsChecker.formatEnumName()` naive title-cased every word, producing labels like `"Shades Of Mortton"` instead of `"Shades of Mort'ton"`.

## Fix detail

**#679 — client-thread name resolution**
- `SourceRecommendedItemsChipPanel` now accepts an optional `ClientThread` (nullable for call sites without one, e.g. `ItemDetailPanel`).
- `buildChip` sets tooltip to `"Item <id>"` immediately (EDT-safe), then schedules a `clientThread.invokeLater` to resolve the real name via `getItemComposition` and pushes it back to the EDT via `SwingUtilities.invokeLater`.
- The `lookupName` method and its `AssertionError` workaround are removed.
- `ClientThread` is threaded through `GuidanceBannerView` (new 3-arg constructor; 1-arg and 2-arg overloads kept for compatibility) and `PanelHeaderBuilder.build()`, sourced from the already-injected `clientThread` in `CollectionLogHelperPlugin`.

**#680 — connector-word lowercasing and apostrophe overrides**
- `LOWERCASE_CONNECTORS` set: common English prepositions/articles (`of`, `the`, `and`, `in`, `a`, `an`, `to`, `for`, `at`, `by`, `or`) are lowercased in non-first position.
- `ENUM_NAME_OVERRIDES` map: explicit display-string overrides for names with apostrophes or punctuation that cannot be reconstructed from the enum constant (bootstrapped with `SHADES_OF_MORTTON → "Shades of Mort'ton"`).

## Tests

- `PanelHeaderBuilderTest`: updated to pass `null` for `clientThread` (no RuneLite runtime in unit tests).
- `RequirementsCheckerTest`: two existing assertions corrected for the now-accurate connector-lowercasing behaviour; six new assertions added covering the `Shades of Mort'ton` override, first-word capitalisation, `and` connector, and `Dragon Slayer` regression guard.
- Full suite: 1673 tests, all pass.

### DIFF
--- a/src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java
+++ b/src/main/java/com/collectionloghelper/CollectionLogHelperPlugin.java
@@ -230,7 +230,7 @@ public class CollectionLogHelperPlugin extends Plugin
 
 		panel = new CollectionLogHelperPanel(
 			config, data.getDatabase(), data.getCollectionState(), efficiency.getCalculator(), efficiency.getClueEstimator(),
-			itemManager, requirementsChecker, data.getDataSyncState(), data.getSlayerTaskState(),
+			itemManager, clientThread, requirementsChecker, data.getDataSyncState(), data.getSlayerTaskState(),
 			efficiency.getSlayerStrategyCalculator(), data.getPlayerInventoryState(), data.getPlayerBankState(),
 			efficiency.getDryStreakAnalyzer(),
 			(java.util.function.BiConsumer<CollectionLogSource, Integer>) this::activateGuidance,
--- a/src/main/java/com/collectionloghelper/data/RequirementsChecker.java
+++ b/src/main/java/com/collectionloghelper/data/RequirementsChecker.java
@@ -429,8 +429,53 @@ public class RequirementsChecker
 		return unmet;
 	}
 
+	/**
+	 * Overrides for quest/source enum names that naive title-casing mis-formats.
+	 *
+	 * <p>Covers two classes of problem:
+	 * <ul>
+	 *   <li>Connector words ("of", "the", "and", "in", "a", "to") that should be
+	 *       lowercase in the middle of a name.</li>
+	 *   <li>Names containing apostrophes or other punctuation that is lost when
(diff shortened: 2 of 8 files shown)


cha-ndler (commented):
## Review — fix(ui): chip-name threading (#679) + Requirements label formatting (#680) (wave 15c)

Verdict: **Approve**. CI green, mergeable. 1673 tests pass.

- **#679 (threading) — correct approach.** The chip panel no longer calls `ItemManager.getItemComposition` on the EDT. It sets `"Item <id>"` immediately, then `clientThread.invokeLater` resolves the real name on the client thread and pushes it back via `SwingUtilities.invokeLater`. The old `lookupName` + its ineffective `catch (RuntimeException)` (which never caught the `AssertionError`) is removed entirely — not papered over with a wider catch, as the issue required. `ClientThread` threaded through new constructor overloads with 1-/2-arg back-compat preserved; `ItemDetailPanel` degrades gracefully. Sound.
- **#680 (formatting) — correct + tested.** `formatEnumName` now keeps connector words (`of/the/and/in/...`) lowercase in non-first position and adds an `ENUM_NAME_OVERRIDES` entry `SHADES_OF_MORTTON -> "Shades of Mort'ton"` for apostrophe names. 6 new unit tests; 2 stale assertions corrected.
- **NOTE (LOW)** — `ENUM_NAME_OVERRIDES` is hand-maintained: a future apostrophe quest not matching the `_S_` split needs a new entry. Acceptable; the connector-lowercasing covers the common case generically.
- **NOTE** — issue #680's point 2 (source listing its own quest as a requirement) was correctly left out of scope.

PR #687, opened 2026-05-25
fix(data): correct Runecrafting (Fire Runes) skill enum to RUNECRAFT

### LINKED ISSUE #678: Data: invalid skill enum "RUNECRAFTING" (should be RUNECRAFT) silently drops a level requirement
## Summary

`drop_rates.json` uses the skill name `"RUNECRAFTING"`, but RuneLite's `net.runelite.api.Skill` enum constant is `RUNECRAFT` (no `-ING`). When `RequirementsChecker` resolves the requirement via `Skill.valueOf("RUNECRAFTING")` it throws `IllegalArgumentException`, the requirement is silently skipped, and a warning is logged.

## Impact

- The Runecrafting level requirement on the affected source is **never enforced** — the source is treated as if it has no Runecrafting requirement (effectively always "met").
- Log spam: because the accessibility scan runs roughly once per game tick during activity, this single bad row produced **924 `WARN c.c.data.RequirementsChecker - Unknown skill enum: RUNECRAFTING`** lines in one short play session.

## Location

- Data: `src/main/resources/com/collectionloghelper/drop_rates.json` line ~29610 — source `"Runecrafting (Fire Runes)"`, alternative `"Rift guardian"`, requirement `"skill": "RUNECRAFTING"`.
- Code that warns: `RequirementsChecker` skill-row resolution (`Skill.valueOf`).

## Fix

Change `"RUNECRAFTING"` to `"RUNECRAFT"` in `drop_rates.json`. Grep the whole file for any other non-enum skill names while here. Run `validate_drop_rates` + regression suite.

cha-ndler:
## Problem
The `Runecrafting (Fire Runes)` source's skill requirement used `"skill": "RUNECRAFTING"`. The RuneLite `Skill` enum constant is `RUNECRAFT` (no -ING), so the value would fail enum resolution silently and the level-14 gate would never apply.

## Fix
`RUNECRAFTING` -> `RUNECRAFT`. Single occurrence. Closes #678.

## Validation
JSON parses; full `./gradlew test` green; ASCII-clean.

### DIFF
--- a/src/main/resources/com/collectionloghelper/drop_rates.json
+++ b/src/main/resources/com/collectionloghelper/drop_rates.json
@@ -29607,7 +29607,7 @@
     "requirements": {
       "skills": [
         {
-          "skill": "RUNECRAFTING",
+          "skill": "RUNECRAFT",
           "level": 14
         }
       ]


cha-ndler (commented):
## Review — fix(data): Runecrafting (Fire Runes) skill enum -> RUNECRAFT (#678)

Verdict: **Approve**. CI running/green, mergeable.

- **VERIFIED** — `RUNECRAFTING` is not a RuneLite `Skill` enum member; the constant is `RUNECRAFT`. The old value would fail enum resolution silently (same class of bug as the wave-14 quest defects), leaving the level-14 gate inert. Single occurrence corrected. Trivial, correct, closes #678.

PR #696, opened 2026-05-26
feat(data): add questMilestones + skillCapePerks requirement fields

cha-ndler:
## Summary

Foundation PR that extends the `SourceRequirements` schema with two additive `@Nullable` detector fields and wires their evaluation in `RequirementsChecker`. No data is wired in this PR — the D5-G/H data-wiring PRs stack on this.

## New fields and exact semantics

**`questMilestones`** (`List<String>`) — RuneLite `Quest` enum constant names where being STARTED is enough. Looser than the existing `quests` field, which requires the quest FINISHED. Evaluated AND-wise: a milestone is satisfied when `Quest.valueOf(name).getState(client)` is `IN_PROGRESS` or `FINISHED` (i.e. not `NOT_STARTED`). Unknown enum names fail closed (logged, skipped), mirroring the `quests` / `pohTeleports` resolution.

**`skillCapePerks`** (`List<String>`) — RuneLite `Skill` enum constant names. A skill-cape travel perk is available only to a player who owns the cape, and owning the cape requires level 99 in that skill. This field uses the level-99 threshold as a proxy for cape ownership (the necessary condition, not proof the cape is worn). Evaluated AND-wise: satisfied when `client.getRealSkillLevel(Skill.valueOf(name)) >= 99`. Unknown enum names fail closed.

Both fields AND-combine with the existing clauses and default to `null`, so all 225 existing sources and all existing JSON parse unchanged (null = no constraint).

## Callsites fixed

`SourceRequirements` is a Lombok `@Value` (generated all-args constructor). The two new fields move it from 7 to 9 positional args. Updated **all 9 affected test callsite locations** (single- and multi-line forms) across 9 test files. No reflective `getDeclaredConstructor` lookups target `SourceRequirements`, so none needed touching.

## Evaluator wiring

Both clauses added to `RequirementsChecker.checkRequirementsDetail(...)`, mirroring the existing `quests` (Quest.valueOf + getState fail-closed) and skill-level (`getRealSkillLevel` + `Skill.valueOf` fail-closed) evaluation shapes.

## Tests added

- **`D5SchemaMigrationTest`** — migration guard mirroring `InventoryItemsAnyMigrationTest`. Walks the whole database (top-level reqs, conditional-alternative reqs flat + nested, waypoint reqs) and asserts both new fields are null EVERYWHERE (empty allowlist). Guards that the fields stay dormant until D5-G/H populate them.
- **`RequirementsCheckerD5Test`** — positive/negative evaluation: questMilestones passes IN_PROGRESS and FINISHED, fails NOT_STARTED; skillCapePerks passes at level >= 99, fails below; both fail closed on unknown enum names.
- **`SourceRequirementsD5GsonTest`** — Gson serialise/deserialise coverage (absent/null/empty/single/multi shapes, full nine-field round-trip, equals).

## Test plan

- [x] `./gradlew test` — BUILD SUCCESSFUL, 1711 tests, 0 failures, 0 errors.
- [x] No changes to `drop_rates.json`.

## Follow-up

D5-G/H data wiring stacks on this foundation.

### DIFF
--- a/src/main/java/com/collectionloghelper/data/RequirementsChecker.java
+++ b/src/main/java/com/collectionloghelper/data/RequirementsChecker.java
@@ -442,7 +442,10 @@ public class RequirementsChecker
 				}
 				catch (IllegalArgumentException e)
 				{
+					// Fail closed: an unverifiable requirement must not silently
+					// promise a shortcut the player may not qualify for.
 					log.warn("Unknown quest enum: {}", questName);
+					unmet.add("Quest started: " + formatEnumName(questName));
 				}
 			}
 		}
@@ -463,7 +466,10 @@ public class RequirementsChecker
 				}
 				catch (IllegalArgumentException e)
 				{
+					// Fail closed: an unverifiable requirement must not silently
+					// promise a shortcut the player may not qualify for.
 					log.warn("Unknown skill enum: {}", skillName);
+					unmet.add("Skill cape: " + formatEnumName(skillName));
 				}
 			}
 		}
--- a/src/main/java/com/collectionloghelper/data/RequirementsChecker.java
+++ b/src/main/java/com/collectionloghelper/data/RequirementsChecker.java
@@ -426,6 +426,48 @@ public class RequirementsChecker
 			}
 		}
 
+		if (requirements.getQuestMilestones() != null)
+		{
+			for (String questName : requirements.getQuestMilestones())
+			{
+				try
+				{
+					Quest quest = Quest.valueOf(questName);
+					// Looser than the quests field: STARTED is enough. Met when
+					// IN_PROGRESS or FINISHED, i.e. anything other than NOT_STARTED.
+					if (quest.getState(client) == QuestState.NOT_STARTED)
+					{
+						unmet.add("Quest started: " + formatEnumName(questName));
+					}
+				}
+				catch (IllegalArgumentException e)
+				{
+					log.warn("Unknown quest enum: {}", questName);
+				}
+			}
+		}
+
+		if (requirements.getSkillCapePerks() != null)
+		{
+			for (String skillName : requirements.getSkillCapePerks())
+			{
+				try
+				{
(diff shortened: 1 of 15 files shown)


cha-ndler (commented):
## Review — D5 schema extension: questMilestones + skillCapePerks (wave 20b)

Verdict: **Approve** (after fail-closed hardening). Foundation PR; D5-G/H data wiring stacks on this.

- **Schema** — two `@Nullable List<String>` fields on `SourceRequirements` (Lombok `@Value`), default null so all 225 sources + existing JSON parse unchanged. 9 `new SourceRequirements(...)` test callsites updated (7->9 args); no reflective `getDeclaredConstructor` lookups target the type. Gson round-trip test added.
- **Semantics (as specified)** — `questMilestones`: met when `Quest.getState` is IN_PROGRESS or FINISHED (looser than `quests`' FINISHED, the "FINISHED is over-restrictive" case). `skillCapePerks`: met when `getRealSkillLevel >= 99` (level-99 proxy for cape ownership, documented in javadoc). Both AND-semantic, mirroring siblings.
- **HARDENED (HIGH, fixed in review)** — the worker's first cut failed OPEN on an unrecognized enum (log-and-skip -> vacuous pass), which for a guidance tool means a typo'd constant promises a shortcut to *every* player. Flipped both `catch (IllegalArgumentException)` clauses to fail CLOSED — mirror the `pohTeleports` clause: unknown enum now adds an unmet entry. Unit tests `*_unknownQuestMilestoneEnum_isFalse` / `*_unknownSkillCapePerkEnum_isFalse` updated to assert the safer behavior. (Pre-existing `quests` fail-open behavior left untouched — out of scope.)
- **Migration guard** — `D5SchemaMigrationTest` walks top-level + conditional-alt (flat+nested) + waypoint reqs, asserts both fields null everywhere (empty allowlist), so the fields stay dormant until G/H intentionally populate them. Full suite 1711+ tests green.

PR #700, opened 2026-05-26
feat(slayer): context-gate strategy panel + fix task-only source filtering

cha-ndler:
## Summary

Gates the slayer strategy advisor to its relevant context and fixes a data-accuracy bug in task-only source filtering.

> **Stacks on #699.** Base is `feat/panel-header-sync`; it will be retargeted to `master` after #699 merges. Review only the slayer/shared changes here.

### Gating
- The strategy advisor now renders **only** in Category Focus mode with the **Slayer** category selected, instead of surfacing globally at the top of the panel. The panel's rebuild path computes `isSlayerContext()` and passes it into `SlayerStrategyView.refresh(boolean)`.

### Data-accuracy fix
- `SlayerStrategyCalculator` now respects `isTaskOnlySource()`: sources obtainable only while on a matching slayer task are no longer counted toward strategy recommendations outside that task.
- New regression test `SlayerStrategyTaskOnlyFilterTest` locks in the filter behavior.

### Icon
- Adds a drawn chevron glyph (`StatusIcons.chevronRight` / `chevronDown`) for the collapsible strategy-section toggle, replacing the ASCII triangle. `SlayerStrategyViewTest` updated to icon-based assertions.

The preview harness gains a `slayer-category-focus` scenario exercising the gated render path.

## Test plan
- [x] `./gradlew compileJava compileTestJava` clean
- [x] `./gradlew test` green (1716 tests, BUILD SUCCESSFUL)
- [x] New `SlayerStrategyTaskOnlyFilterTest` passes

### DIFF
--- a/src/main/java/com/collectionloghelper/efficiency/SlayerStrategyCalculator.java
+++ b/src/main/java/com/collectionloghelper/efficiency/SlayerStrategyCalculator.java
@@ -238,7 +238,14 @@ public class SlayerStrategyCalculator
 
 	/**
 	 * Returns true if the given Slayer creature maps to any collection log
-	 * source that has at least one missing item.
+	 * source that has at least one missing item AND whose unique drops genuinely
+	 * require an active Slayer task.
+	 * <p>
+	 * Sources that are freely farmable off-task (per
+	 * {@link SlayerCreatureDatabase#isTaskOnlySource(String)}) are excluded: they
+	 * provide no reason to keep this task assigned or to choose a master for it,
+	 * so they must not influence the block list, best-master recommendation, or
+	 * the useful-task probability that drives master comparison.
 	 */
 	private boolean hasAnyMissingItems(String creatureName)
 	{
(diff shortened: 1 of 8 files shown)


cha-ndler (commented):
## Review — context-gate slayer strategy + task-only filtering fix

**Conflicts resolved.** The base PR was squash-merged to master, leaving this
branch carrying the now-redundant header commit. Rebased `--onto master` (dropping
that commit) so the branch is a single slayer-only commit; diff against master is
exactly the 8 intended files. `compileJava`/`compileTestJava` + full `test` suite
pass. State is MERGEABLE / CLEAN.

**Data-accuracy fix (the important part): correct and well-placed.** The
`isTaskOnlySource()` guard sits at the single `hasAnyMissingItems` chokepoint, so
it consistently flows into the block list, best-master recommendation, and
master-comparison without duplicating the rule. Per-source granularity is right:
a creature mapping to both an off-task source and a task-only boss variant counts
only via the task-only one. The new `SlayerStrategyTaskOnlyFilterTest` covers both
the block-list and best-master paths with realistic off-task (gargoyle) vs
task-only (Cerberus) fixtures.

No CRITICAL/HIGH findings. Two scope notes:

**NOTE:** the panel gate is Category Focus + SLAYER category only. The
"slayer item selected in another mode" path from the original idea is not wired;
acceptable for this change, can follow up if desired.

**NOTE:** the current-task readout (`getUsefulSourcesForCreature` /
`getMissingItemsForCreature`) is intentionally left unfiltered, since the player is
already on that task and listing every useful source there is correct.

**Verdict:** merge-ready. Squash; base master.

Issue #719, opened 2026-05-26
Guidance: derive current step from player state (location + inventory + clog completion)


cha-ndler:
## Guidance: derive the current step from player STATE (location + inventory + collection-log completion)

Guidance currently advances purely event-driven, which causes:
- Getting **stuck on the final step** of looping activities. At Shades of Mort'ton, when the player runs out of shade keys, guidance stays on "Step 5/5: Open chests matching your shade keys" instead of resetting to the bank step to restock (observed live).
- No **mid-activity re-sync**: activating guidance while already mid-activity doesn't place the player at the right step.

### Desired logic
On guidance activation AND on relevant state changes, derive the correct step from the player's STATE:
- If the player is already in a step's **area** and has that step's **required items**, place them at that step (e.g. already in the Shades catacomb holding keys -> the "open chests" step).
- At the **end**: check whether the target collection-log item is obtained. If NOT obtained and the player is **out of the current step's required items**, reset to the earliest step whose required items the player needs (e.g. back to the bank step when they have no keys / shade remains / pyre logs).

This state-derivation (location + inventory + clog-completion) fixes the loop-back/ending and enables resuming guidance mid-activity. Relates to #707 (recurring-item highlight) and the loopBackToStep logic; likely the core remaining guidance-engine work.

PR #725, opened 2026-05-27
feat(guidance): keep recurring-item highlight active past the first pickup

### LINKED ISSUE #707: Keep highlighting recurring step items after the first is obtained
## Enhancement / open question: keep highlighting recurring step items after the first is obtained

Today the in-game highlight is driven by the current step's requirement and effectively highlights only the first matching item/object. After the first key is obtained, subsequent keys of the same type are no longer highlighted, even though the player still needs to collect/use more of them on the same step.

Notes:
- Possible approach: continue highlighting the relevant key objects across the early steps (e.g. steps 1 and 2) rather than de-highlighting after the first pickup.
- May be an acceptable gap if there is no clean signal to keep the highlight active; open to ideas on intelligently re-highlighting recurring items once the step has progressed.

Relates to the required-item highlight enhancement.

cha-ndler:
## Summary

Keeps the in-game required-item highlight active past the first pickup for
looping / restock activities.

In a looping/restock activity (e.g. Shades of Mort'ton) the player gathers
many of a recurring consumable per run. The gather step highlights the keys
on the ground and declares `skipIfHasAnyItemIds`, so obtaining the **first**
key auto-advanced off the step in `onInventoryChanged` and cleared the key
highlight while more keys still needed collecting.

## Mechanism

Reuses the existing looping/restock signal — a sequence is a recurring
activity when any of its steps declares `restockIfMissingAllItemIds` (added
in #719/#721). When that signal is present, the `skipIfHasAnyItemIds`
auto-advance is suppressed, so the gather step (and its ground/item highlight)
stays active until the player advances manually (Next/Skip) or the loop
naturally re-enters it.

- New private helper `GuidanceSequencer.isRecurringGatherSequence()` scans the
  active steps for the restock-fuel signal.
- The single skip-auto-advance branch in `onInventoryChanged` now gates on
  `!isRecurringGatherSequence()`.
- **No new schema field** — the existing `restockIfMissingAllItemIds` already
  marks a sequence as a consume-many loop.

## Test plan

- [x] New unit test `recurringGatherStepKeepsHighlightAfterFirstPickup` in
  `GuidanceSequencerRestockTest`: a looping sequence's gather step stays active
  with its highlight target unchanged after one key enters inventory.
- [x] Existing restock/latch tests unchanged and green.
- [x] `./gradlew build` green (tests + coverage + drop-rate lint).

Closes #707

### DIFF
--- a/src/main/java/com/collectionloghelper/guidance/GuidanceSequencer.java
+++ b/src/main/java/com/collectionloghelper/guidance/GuidanceSequencer.java
@@ -562,9 +562,18 @@ public class GuidanceSequencer
 
 		GuidanceStep step = getRawCurrentStep();
 
-		// Auto-advance if skipIfHasAnyItemIds is now satisfied (e.g., key just entered inventory)
+		// Auto-advance if skipIfHasAnyItemIds is now satisfied (e.g., key just entered inventory).
+		//
+		// Exception (#707): in a looping / restock activity the player must keep
+		// gathering many of a recurring item (e.g. several Shade keys per catacomb
+		// run) before moving on. Auto-advancing off the gather step after the FIRST
+		// pickup clears its in-game highlight prematurely, so the player loses sight
+		// of the remaining keys to collect. While the sequence is a looping/restock
+		// activity, suppress the skip auto-advance so the gather step — and its
+		// highlight — stays active until the player advances manually (Next/Skip) or
+		// the loop naturally re-enters it.
 		CompletionChecker.SkipIfHasAnyResult skip = completionChecker.evaluateSkipIfHasAny(step);
-		if (skip.matched())
+		if (skip.matched() && !isRecurringGatherSequence())
 		{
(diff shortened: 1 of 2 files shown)


cha-ndler (commented):
Review (orchestrator pass):

MEDIUM — The suppression in onInventoryChanged is sequence-wide: isRecurringGatherSequence() returns true if ANY step declares restockIfMissingAllItemIds, which then disables skipIfHasAnyItemIds auto-advance for EVERY step in that sequence, not only the gather step. For a looping/restock activity that also has a legitimate 'skip prep step if you already hold X' step, that auto-advance is lost. Low real-world risk today (few looping sequences, and the restock latch handles re-entry), acceptable as a minimal fix — but worth a follow-up if a future looping source needs a mid-sequence skip.
LOW — Mechanism reuses the existing restock signal instead of adding a schema field; correct call. Test constructs the step via the full positional constructor with trailing nulls (no other callsites touched). Build green.

No blocking issues. Approving.

PR #726, opened 2026-05-27
fix(guidance): keep overlay item availability in sync with the side panel

### LINKED ISSUE #717: In-game overlay item availability (e.g. Oak pyre logs) diverges from the side panel
## In-game overlay item-requirements don't reflect bank/inventory availability when the side panel does

Live testing (Shades of Mort'ton): the in-game guidance overlay's "Item requirements" list showed "Oak pyre logs" with no held/(bank) annotation, while the side panel correctly detected the player HAS them (in bank). Tinderbox / Phrin remains were annotated "(bank)" correctly in the overlay; Oak pyre logs was missed.

The side panel (RequiredItemResolver / RequiredItemDisplay) and the in-game overlay are expected to share availability resolution, yet they diverge here. Needs root-cause: whether the overlay resolves a stale/different step context, a tiered-item (pyre-log tier) ID mismatch, or a different inventory/bank source than the panel. Pinpoint the divergence before fixing.

cha-ndler:
## Root cause

For a tiered/upgradeable required item (the report's example is Oak pyre logs in Shades of Mort'ton), the in-game overlay's "Item requirements" availability disagreed with the side panel: the panel showed the item as held in the bank while the overlay showed it as missing.

Both surfaces resolve availability through the same `RequiredItemResolver`, but they are pushed on different cadences:

- The **overlay** receives its required-item availability only inside `OverlayStepApplier.apply`, which runs from `GuidanceOverlayCoordinator.onStepChanged` **only on a genuine step change**. On a same-step re-notify (fired by `GuidanceSequencer.onInventoryChanged` on every inventory change while guidance is active — and a bank scan/withdrawal triggers these) the apply path is intentionally skipped to avoid a visual blink (#683).
- The **side panel** (`StepChangeHandler.handle`) re-resolves availability on **every** notify, including same-step re-notifies.

So once the player's bank contents became known after step entry, the panel's row flipped to `IN_BANK` but the overlay kept its stale step-entry snapshot (`MISSING`). Items already known at step entry (Tinderbox, Phrin remains) looked correct, masking the divergence to just the later-resolved tier item.

## Fix

In `onStepChanged`, on a same-step re-notify, re-resolve the step's required items through the **same** `RequiredItemResolver` the panel uses and push them to the overlay (deferred to the client thread, mirroring the existing `OverlayStepApplier` / `StepChangeHandler` pattern). Both surfaces now share one availability source and agree.

Single-file change plus a regression test; the genuine-step-change overlay path is untouched, so the #683 anti-blink behaviour is preserved.

## Test plan

- [x] `GuidanceOverlayCoordinatorRequiredItemSyncTest` reproduces the divergence: step-entry resolve returns `MISSING`, a bank scan flips the resolver to `IN_BANK`, and a same-step re-notify must re-push the refreshed `IN_BANK` availability to the overlay (matching the panel).
- [x] `./gradlew build` green (full suite + coverage + data lint).
- [x] Existing `GuidanceOverlayCoordinatorStepGuardTest` still passes (genuine-change vs re-notify apply counts unchanged; #683 anti-blink preserved).

Fixes #717

### DIFF
--- a/src/main/java/com/collectionloghelper/guidance/GuidanceOverlayCoordinator.java
+++ b/src/main/java/com/collectionloghelper/guidance/GuidanceOverlayCoordinator.java
@@ -624,9 +624,40 @@ public class GuidanceOverlayCoordinator
 			applyStepToOverlays(step, applySourceName, applySource);
 			npcTrackerHelper.scanForTrackedNpc(step, activeTargetItemId);
 		}
+		else
+		{
+			// Same-step re-notify (fired on every inventory change while guidance is
+			// active, e.g. a bank withdrawal during Shades of Mort'ton). The overlay
+			// clear + re-apply above is skipped to avoid a visual blink (#683), but
+			// that path is also the only one that re-pushes the overlay's required-item
+			// availability via OverlayStepApplier. Without re-resolving here the overlay
+			// keeps a stale availability snapshot from step entry while the side panel
+			// (StepChangeHandler, below) re-resolves on every re-notify -- so a freshly
+			// scanned bank item shows "(bank)" in the panel but not the overlay (#717).
+			// Re-resolve here through the same RequiredItemResolver the panel uses so
+			// both surfaces share one availability source.
+			refreshOverlayRequiredItems(step);
+		}
 
 		stepChangeHandler.handle(step,
 			new StepChangeHandler.Input(activeInfoBox, panel, activeTargetItemId));
+	}
+
+	/**
+	 * Re-resolves the given step's required-item availability and pushes it to the
+	 * in-game overlay, keeping the overlay's "Item requirements" list in sync with the
+	 * side panel on same-step re-notifies (#717).
+	 *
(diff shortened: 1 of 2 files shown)


cha-ndler (commented):
Review (orchestrator pass):

LOW — Root cause is precise and well-documented: the same-step re-notify path skips the OverlayStepApplier apply (to preserve the #683 anti-blink behaviour), which is the only path that re-pushes overlay required-item availability, so the overlay held a stale step-entry snapshot while the panel re-resolved every notify (#717). Fix re-resolves through the same RequiredItemResolver on the client thread (getItemComposition is client-thread-only, #388) and leaves the genuine-step-change path untouched.
LOW — refreshOverlayRequiredItems runs resolve() on every same-step re-notify, but that matches the panel's existing cadence and getItemComposition is cached — no new perf concern.

Thorough regression test reproduces the MISSING->IN_BANK divergence. Build green. No blocking issues. Approving.

PR #727, opened 2026-05-27
feat(guidance): derive starting step from player state on activation

### LINKED ISSUE #719: Guidance: derive current step from player state (location + inventory + clog completion)

cha-ndler:
## Summary

Implements the remaining half of #719: **mid-activity re-sync on guidance activation**. When the player activates guidance while already mid-activity, the sequencer now lands them on the correct step derived from their current STATE instead of always starting at step 1.

> Note: the loop-back / restock-on-depletion half of #719 was already shipped in #721 (`restockIfMissingAllItemIds` + restock latch). This PR does not touch it — the new derivation runs *before* the existing depletion check and composes with it.

## Design

On activation, after the forward skip-chain runs, scan forward for the **furthest-along** step whose:
- **area is positively confirmed** — player within the step's `completionDistance` of its `worldX/worldY/worldPlane` tile, OR inside its `completionZone` — AND
- **effective required items** are fully held (reuses the existing per-item-aware required-item predicate).

If a later qualifying step is found, jump there; otherwise leave the skip-chain result unchanged. A step whose location is unconfirmable (no tile and no zone, or unknown player location) is **never** a forward target — wrong-forward placement is worse than starting at step 1. No new schema field: derivation uses only existing step data. Wired into both `startSequence` and `syncToCurrentState` (the Sync button benefits too).

## Files changed

- `src/main/java/com/collectionloghelper/guidance/GuidanceSequencer.java` — `advanceToFurthestSatisfiedState()` + `playerIsConfirmablyInStepArea()`, called from `startSequence` and `syncToCurrentState`.
- `src/test/java/com/collectionloghelper/guidance/GuidanceSequencerStateDerivationTest.java` — positive case (in catacomb zone holding keys → open-chests step), safety case (ambiguous state → step 1), and an in-area-but-missing-items case (no jump).

## In-game validation TODO

Cannot be validated from the dev environment — manual repro needed:
- [ ] Activate guidance at Shades of Mort'ton while standing in the catacomb holding shade keys → expect to land on the open-chests step, not the bank/travel steps.
- [ ] Activate the same source while standing at the bank with no keys → expect step 1 (bank) unchanged.
- [ ] Stand in the catacomb with NO keys, activate → expect step 1 (no forward jump, items missing).
- [ ] Confirm the Sync button performs the same state-derived jump mid-session.
- [ ] Spot-check a non-looping multi-step source (e.g. a zone-gated source) to confirm no spurious forward jumps when the player is between areas.

Closes #719

### DIFF
--- a/src/main/java/com/collectionloghelper/guidance/GuidanceSequencer.java
+++ b/src/main/java/com/collectionloghelper/guidance/GuidanceSequencer.java
@@ -184,6 +184,11 @@ public class GuidanceSequencer
 		// Skip any steps whose conditions are already met
 		skipSatisfiedSteps();
 
+		// Mid-activity re-sync (#719): if the player is already standing in a
+		// later step's confirmable area AND holds that step's required items,
+		// jump them forward to it instead of starting at step 1.
+		advanceToFurthestSatisfiedState();
+
 		// Mid-activity activation while already depleted (e.g. in the Shades
 		// catacombs with no keys and no remains) — park on the restock step.
 		checkDepletionAndMaybeReset();
@@ -379,6 +384,10 @@ public class GuidanceSequencer
 				activeSource != null ? activeSource.getName() : "?");
 			return;
 		}
+
+		// State-derived forward jump (#719): land on the furthest step whose
+		// area + required items the player demonstrably satisfies.
+		advanceToFurthestSatisfiedState();
 
 		// If the player synced while depleted, park on the restock step.
 		checkDepletionAndMaybeReset();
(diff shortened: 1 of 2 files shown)


cha-ndler (commented):
Review (orchestrator pass):

LOW — Design is appropriately conservative for a behavior-risky engine change. advanceToFurthestSatisfiedState() only jumps FORWARD, and only to a step whose area is POSITIVELY confirmed (inside completionZone, or within completionDistance of the worldX/Y/plane tile) AND whose required items are all held. Unconfirmable steps (no tile, no zone, or unknown player location) are never jump targets, so the fallback is the current step-1 behaviour. Reuses the existing playerHasAllRequiredItems predicate + getZone/completionDistance — no new schema field, no parallel logic. Ordering (skipSatisfiedSteps -> advanceToFurthestSatisfiedState -> checkDepletionAndMaybeReset) is correct.

MEDIUM (validation, not code) — The derivation also runs inside syncToCurrentState(), so mid-guidance the step can jump forward whenever the player walks into a later step's confirmed area while holding its items. That matches the issue's 'on relevant state changes' intent and is forward-only (depletion/restock handles the reset path), but it's the riskiest behaviour and must be confirmed in-game before this is considered user-verified. The PR's in-game-validation TODO covers the Shades-of-Mort'ton repro — please also validate a multi-step source where consecutive steps share nearby coordinates (e.g. a same-tile multi-phase boss) to confirm it doesn't over-jump.

Tests: positive (catacomb + keys -> open-chests), safety (ambiguous -> step 1), and in-area-but-missing-items (-> no jump). Good coverage of the safety contract. No blocking issues — approving pending green CI. Half 1 (#721) correctly left untouched.

PR #735, opened 2026-06-04
test(data): add global source-invariant audit across all sources

cha-ndler:
## Gap

The repo has many per-batch and per-category regression/audit tests, but they assert only over hard-coded name allowlists (e.g. `SoloBossesDeepGuidanceAuditTest` iterates a fixed `SOLO_BOSS_SOURCES` list). A source that is not enumerated in any list can silently lose its `guidanceSteps`, coordinates, or kill time and no test catches it.

## What this adds

A single global guard, `SourceInvariantAuditTest`, that loads the production database and iterates **every** source from `DropRateDatabase.getAllSources()` (no name allowlist). It accumulates all violations and fails the build if any source regresses below baseline authoring invariants.

## Invariants asserted (per source)

1. `getName()` is non-null and non-blank.
2. `getGuidanceSteps()` is non-null and non-empty.
3. Step count `>= 2`, except an allowlist of catch-all pseudo-sources (`Miscellaneous`, `Random Events`) which require `>= 1`.
4. `getWorldX() > 0` and `getWorldY() > 0` (primary location); `getWorldPlane() >= 0`.
5. `getKillTimeSeconds() > 0` OR `getIronKillTimeSeconds() > 0`.
6. Each step's `getDescription()` is non-null and non-blank.

## Single-step allowlist

`Miscellaneous` and `Random Events` are catch-all pseudo-sources, not real activities, so they legitimately carry one informational step. All other sources must have at least two steps.

## Test plan

- [x] `./gradlew build` succeeds.
- [x] `./gradlew test` passes — 1753 tests, 0 failures, 0 errors, including the new `SourceInvariantAuditTest`.
- [x] Test-only change; no production or `drop_rates.json` modification.

### DIFF
--- a/src/test/java/com/collectionloghelper/data/SourceInvariantAuditTest.java
+++ b/src/test/java/com/collectionloghelper/data/SourceInvariantAuditTest.java
@@ -1,0 +1,154 @@
+/*
+ * Copyright (c) 2025, cha-ndler
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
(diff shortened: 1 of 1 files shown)


cha-ndler (commented):
## Orchestrator review — PR #735 (global source-invariant audit)

**Verdict: APPROVE / merge.** No CRITICAL or HIGH findings.

### Validation gate (all green)
- `./gradlew build` — SUCCESS
- `./gradlew test` — 1753 tests (+1), 0 failures; the new `SourceInvariantAuditTest` passes as authored
- Diff is **test-only** — one new file, no production/data change

### Correctness (severity: NOTE)
- Closes a real gap: the per-category `*DeepGuidanceAuditTest` guards iterate hard-coded name allowlists, so a source absent from every list could silently lose guidanceSteps/coords/kill-time. This guard iterates `getAllSources()` and covers all 225.
- Invariants match the empirically-verified baseline (0 sources currently violate any): name non-blank · ≥2 steps (allowlist `Miscellaneous`/`Random Events` at ≥1, the two catch-all pseudo-sources) · worldX>0 && worldY>0, worldPlane≥0 · killTimeSeconds>0 || ironKillTimeSeconds>0 · every step description non-blank.
- Accumulate-then-assert with a joined violation message — a future regression reports every offending source at once, not just the first. Matches the sibling regression-test idiom.
- Allowlist is a documented `Set.of(...)` constant, narrowly scoped — no risk of masking real regressions on other sources.

### Note (severity: INFO)
Threshold choices are coupled to current data; if a legitimate new single-step pseudo-source is ever added, it must be added to `SINGLE_STEP_ALLOWLIST` with a comment — which is the correct, explicit failure mode.

PR #1058, opened 2026-06-29
fix(guidance): hint on MANUAL loop / recurring-gather steps so they do not read as stuck (#822)

### LINKED ISSUE #822: Recurring-gather MANUAL steps read as 'stuck' - add a 'keep gathering, press Next' hint
## Observed (live, 2026-06-11)

On the Shades of Mort'ton gather step (MANUAL, "Pick up your shade key"), the recurring-gather suppression (#707/#725) correctly keeps the step active past the first key pickup instead of auto-advancing — so the highlight persists for repeat keys. **This is working as designed (T2.11).**

But from the player's seat it reads as **stuck**: they picked up a key, the step did not advance, they deposited items and teleported expecting progression, and it stayed on the same step. Nothing on-screen tells them the step is *intentionally* waiting and that they should press **Next** when done gathering (or that the loop will re-enter it).

## Ask

On a recurring-gather step (a sequence where `isRecurringGatherSequence()` is true and the active step's `skipIfHasAnyItemIds` auto-advance is suppressed), surface a small hint so the suppression doesn't read as a hang — e.g. a side-panel / overlay note like *"Keep gathering — press Next when done"* on the gather step.

## Scope

Applies to every recurring-gather sequence, not just Shades (any source with `restockIfMissingAllItemIds` + a `skipIfHasAnyItemIds` gather step). A one-line hint keyed off the same `isRecurringGatherSequence()` condition covers them all.

## Related

- Sibling data bug: the Shades gather step also highlights a "reward pillar" instead of the key, compounding the "stuck" impression (separate issue).

cha-ndler:
Fixes #822.

Recurring-gather and MANUAL loop-terminal steps intentionally suppress auto-advance, which reads as "stuck". Adds `GuidanceSequencer.getActiveStepHint()` (one source of truth) returning a short cue for intentional wait steps:
- recurring-gather (`isRecurringGatherSequence()`): "Keep gathering - press Next when done"
- plain MANUAL loop terminal (`loopBackToStep`/`loopCount`): "Repeat until done - press Next to continue, or Stop to finish"
- one-off MANUAL / event-driven: `null` (unchanged)

`StepChangeHandler` appends it to the InfoBox tooltip + side-panel description. Keyed off the loop/recurring condition, so all 14 MANUAL-loop sources benefit, not just Shades (retroactive, per live calibration triage).

**Validation:** new unit test (recurring-gather step surfaces the hint) + full guidance suite green. Live seam validation (driving Wintertodt/Shades and asserting `state.read.stepHint`) to follow on the next dev-client relaunch.

Structural guidance-engine change - no auto-merge; for review.

### DIFF
--- a/src/main/java/com/collectionloghelper/guidance/GuidanceSequencer.java
+++ b/src/main/java/com/collectionloghelper/guidance/GuidanceSequencer.java
@@ -248,6 +248,32 @@ public class GuidanceSequencer
 	}
 
 	/**
+	 * #822: a short user-facing hint for the active step when it is an intentional
+	 * WAIT step — a recurring-gather step (auto-advance deliberately suppressed) or a
+	 * MANUAL loop terminal — so the deliberate pause is not mistaken for a hang.
+	 * Returns {@code null} for a one-off MANUAL step or an event-driven step (those
+	 * advance on their own or on a single Next, so they never read as stuck).
+	 * Keyed off the loop/recurring condition, so every loop source benefits at once.
+	 */
+	public String getActiveStepHint()
+	{
+		GuidanceStep step = getCurrentStep();
+		if (step == null || step.getCompletionCondition() != CompletionCondition.MANUAL)
+		{
+			return null;
+		}
+		if (isRecurringGatherSequence())
+		{
+			return "Keep gathering - press Next when done (the loop re-enters this step).";
+		}
+		if (step.getLoopBackToStep() > 0 && step.getLoopCount() > 0)
+		{
+			return "Repeat until done - press Next to continue, or Stop to finish.";
+		}
+		return null;
+	}
+
+	/**
 	 * Returns the raw current step without bank routing substitution,
 	 * but with conditional alternatives resolved.
 	 */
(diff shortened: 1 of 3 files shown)


cha-ndler:
Live seam validation (dev client, harness v0.6 with state.read.stepHint):

- **Plain MANUAL loop branch — confirmed live.** `guidance.gotoStep` Wintertodt step 2 (the feed/fight MANUAL loop) → `state.read` returns `stepHint: "Repeat until done - press Next to continue, or Stop to finish."`
- **Recurring-gather branch — unit-validated** (`GuidanceSequencerRestockTest.recurringGatherStepSurfacesKeepGatheringHint`). It cannot currently be parked live because the active step on a state-driven gather source (Shades) is re-derived from real inventory/location on `gotoStep`, landing on the non-MANUAL bank step. That re-derive-reads-real-location gap is a separate drive-fidelity item (would need a `setLocation` seam primer); it does not affect the fix, which is keyed off the same condition the unit test exercises.

`getActiveStepHint()` is the single source of truth feeding both the InfoBox tooltip and the side-panel description.

